PR #6, opened 2017-12-12
Inject Raven to the context as $raven

anteriovieira:
We could do something like:

```js
import {User} from './api'
export default {
    // ..
    methods: {
        async fetchUsers() {
            try () {
                this.users = await User.fetchAll()
            } catch (err) {
                this.$raven.captureException(err)
            }
        }
    }
}
```

### DIFF
--- a/lib/templates/sentry-client.js
+++ b/lib/templates/sentry-client.js
@@ -2,7 +2,13 @@ import Vue from 'vue'
 import Raven from 'raven-js'
 import RavenVue from 'raven-js/plugins/vue'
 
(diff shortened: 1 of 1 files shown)


anteriovieira:
CC @DiederikvandenB , @pi0

DiederikvandenB:
Interesting idea. I will look in to this further today or tomorrow. Thanks for your efforts ;).

DiederikvandenB:
Merged + I also updated the raven packages to their newest versions.

PR #18, opened 2018-03-22
Add option.dsn check

### LINKED ISSUE #17: must have a SENTRY_PRIVATE_KEY env, even though SENTRY_DSN exists
This module works well, but one issue I ran into is that I had to have SENTRY_PRIVATE_KEY defined for this to work... even though SENTRY_DSN was defined.

It looks like the culprit is here:
https://github.com/nuxt-community/sentry-module/blob/master/lib/sentry.js#L22

I'm happy to fix, just wanted to make sure there wasn't more to this before doing so.  Thanks!

pauldariye:
Fixes #17 

Based on @DiederikvandenB's solution

### DIFF
--- a/lib/sentry.js
+++ b/lib/sentry.js
@@ -19,7 +19,7 @@ module.exports = function sentry (moduleOptions) {
   }, this.options.sentry, moduleOptions)
 
   // Don't proceed if no keys are provided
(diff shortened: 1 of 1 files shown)


DiederikvandenB:
Thanks for the PR @pauldariye!

PR #25, opened 2018-07-05
use debug module like other nuxt packages

btkostner:
Before:
screenshot from 2018-07-05 16 01 47

After:
screenshot from 2018-07-05 16 02 44

### DIFF
--- a/lib/sentry.js
+++ b/lib/sentry.js
@@ -1,6 +1,8 @@
 const Raven = require('raven')
 const path = require('path')
 
+const debug = require('debug')('nuxt:sentry')
+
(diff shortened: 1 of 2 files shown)


pi0:
Using consola with upcoming 2.0 release is more prefered. Nuxt and Some official modules already migrated too :)

DiederikvandenB:
I'll look in to it after the weekend.

DiederikvandenB:
@btkostner, are you willing to update your PR to use Consola?

btkostner:
@DiederikvandenB sure

PR #28, opened 2018-08-17
feat: Add new option "disabled"

mirucon:
This PR adds a new option "disabled" to completely disable error capturing for both client/server. This might be useful they're not needed in the development env.


DiederikvandenB:
Merged.

PR #29, opened 2018-09-08
Fix checking the option 'disabled'

Zakhse:
Option `disabled` with `true` value is ignored because of bug in the module.
This PR fixes it.


DiederikvandenB:
Thanks. On mobile now, will release to npm later today.

PR #41, opened 2018-11-13
client side ssr on/off option

Valeronlol:
Hey!
Can we provide option to chose ssr mode when register the client plugin?
It's will be realy helpful when we want to use $sentry incide our custom plugins.
For example like this https://www.dropbox.com/s/huzghaz4anju1ls/Screenshot%202018-11-13%2013.31.49.png?dl=0

### DIFF
--- a/lib/sentry.js
+++ b/lib/sentry.js
@@ -8,6 +8,7 @@ module.exports = function sentry (moduleOptions) {
   const options = Object.assign({
     disabled: process.env.SENTRY_DISABLED || false,
     disableClientSide: process.env.SENTRY_DISABLE_CLIENT_SIDE === 'true' || false,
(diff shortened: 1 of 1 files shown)


aldarund:
sentry-client.js use a sentry\browser package that is supposed to be used in browser only. For node env there is @sentry/node which is used here https://github.com/nuxt-community/sentry-module/blob/master/lib/sentry.js

Valeronlol:
Thanks, I saw it, but is there any way to use app.$sentry incide plugins?

manniL:
@Valeronlol Your plugin default export function has access to `ctx` (`ctx.app.$sentry`)

DiederikvandenB:
Thanks @manniL and @aldarund for providing support! @Valeronlol, it seems like your question has been answered, and that this PR is not necessary.

Issue #21, opened 2018-05-04
Ability to disable for specific environments?


acidjazz:
I'd like to not load this for my local env, is there any way i can dynamically load the module and/or specify a config option that disables it and then decide based on an ENV variable?

DiederikvandenB:
There are multiple ways to achieve this. In our opinion, the easiest way to do this is to use `.env` files. Quick example:

`.env.local`
```
SENTRY_DSN=
```

`.env.production`
```
SENTRY_DSN=your sentry dsn
```

Just make sure that your production environment uses the correct env file and you're set.

acidjazz:
@DiederikvandenB great idea! thanks!

PR #46, opened 2018-12-13
Make sentry disabled if it's not in production

nmfzone:
I think it's good to have default disabled value to true when it's not in production environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Will be ignored if `dsn` provided.
 
 ### disabled
 - Type: `Boolean`
-  - Default: `process.env.SENTRY_DISABLED || false`
+  - Default: `process.env.SENTRY_DISABLED || process.env.NODE_ENV !== 'production'`
 
 ### disableClientSide
 - Type: `Boolean`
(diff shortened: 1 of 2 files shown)


DiederikvandenB:
What is wrong with the method described here:
https://github.com/nuxt-community/sentry-module/issues/21#issuecomment-386746094

I do not really see how this PR is an enhancement, but I am open for a discussion.

nmfzone:
@DiederikvandenB Ok, let me explain it. The reason of this PR, it comes from my workflow.
Sometimes I need to test production build on my local machine, coz I need to make sure everything works.

I know it can be easily disabled if no DSN provided, but I think it could be handy if I don't need to touch the env (remove DSN in local, and fill it in production).

So, if I set the DSN, sentry will disabled by default on local, and enabled on production.

DiederikvandenB:
In that case, just run your application locally as such:

`SENTRY_DISABLED=false yarn run [cmd]`

I am going to close this PR as I do not think the proposed changes are desirable. Feel free to maintain your own fork of this repo if you want.

Issue #51, opened 2019-01-29
disabled option does not work


tic40:
### Version

[v2.2.1](https://github.com/sentry-module/releases/tag/v2.2.1)

### Reproduction link

[https://jsfiddle.net/y5sa4ht8/](https://jsfiddle.net/y5sa4ht8/)

### Steps to reproduce

Set the following sentry config in nuxt.config.js.

```
{
  sentry: {
    dsn: 'dsn',
    disabled: true,
    config: {
      environment: process.env.NODE_ENV // 'development'
    }
  }
}
```

### What is expected ?

I expect that sentry-module does not send error reports because disabled option is true.

### What is actually happening?

Though I set `disabled: true`, sentry-module sends error reports.

### Additional comments?

This works well on v2.1.0. I met this on v2.2.1.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c43">#c43</a>)</em></sub></div>

DiederikvandenB:
See #53

PR #53, opened 2019-02-07
Fix: disabled option does not work (#51)

mya-ake:
'return' disappeared at this commit. So it does not work.
https://github.com/nuxt-community/sentry-module/commit/31d3db941ab79cf79d52e30d2986f0fe373a0f8b#diff-e6ee66530234110522eee8b87cee6b9f


tomoyukikashiro:
@DiederikvandenB Plz merge this PR 🙇 @mya-ake thx !

PR #57, opened 2019-02-19
Fix: @sentry/webpack-plugin dependency

edcrampin:
@sentry/webpack-plugin is required as a dependency but is currently listed under devDependencies.


DiederikvandenB:
Can't believe I've missed that. Sorry!

PR #58, opened 2019-02-20
fix: sentry init params

runnez:
It fixes initialization on a server 🚑

```
{
   ...options.dsn,
   ...options.config
}
```
options.dsn is a string. spread causes incorrect options passing into sentry. 
```
{                                                                 
  '0': 'h',
  '1': 't',
  '2': 't',
  ...
  '58': '3',
  environment: 'development'
}
```

### DIFF
--- a/lib/sentry.js
+++ b/lib/sentry.js
@@ -25,8 +25,8 @@ module.exports = function sentry (moduleOptions) {
   }
 
   const options = Object.assign({}, defaults, this.options.sentry, moduleOptions)
-  options.config = Object.assign({}, defaults.config, this.options.sentry.config, moduleOptions.config)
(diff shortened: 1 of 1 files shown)


runnez:
🚑

rchl:
I was fixing other bug and also handled this in #59.

runnez:
@rchl I think, my RP do same things without extending dependencies

rchl:
My fix fixes additional bug though - exception when not using top level settings.

rchl:
Ah, you have added some more changes now. The thing is, your approach is more fragile and will need changes when additional child objects are added. Also you missed some lines that will still crash.

My point is not to compete with your fix though, just create robust solution that has less chance of breaking.

runnez:
omg, you are right! 
I will be happy if any PR merge :)

pi0:
Thanks @runnez for this PR. This should be fixed by #59.

PR #59, opened 2019-02-25
fix: fix module options

rchl:
With no top level 'sentry' object defined in nuxt-config.js, the code
would try to access property of undefined object as
this.options.sentry is undefined. Fixed by using deepmerge package which
both fixes the problem and makes code cleaner.

Also fixed problem with merging dsn property value with the object.
As dsn value is a string, don't use ... on it as that would turn string
into an array of letters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 The module enables error logging through [Sentry](http://sentry.io).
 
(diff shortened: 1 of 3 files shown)


pi0:
Thanks @rchl for this fix and sorry for long delay.

PR #60, opened 2019-03-06
The additional options are called `config`, not `options`

weotch:


pi0:
This fixed by #59. BTW thanks for this PR :heart:

Issue #33, opened 2018-10-30
How to use sentry on vuex action (serverside)


eseceve:
Hi, I trying to use `$sentry` from a vuex action but it's `undefined` when action is called from `asyncData`

`store/index.js`
```javascript
async foo({ commit, state }) {
   const response = await this.$axios.$get('/foo/')
   this.$sentry.captureMessage('Test foo', { extra: { response }})
}
```
`package.json`
```
"@nuxtjs/sentry": "^2.0.0",
"nuxt": "^2.2.0",
```

`nuxt.config.js`
```javascript
  // read dsn from process.env.SENTR_DSN
  sentry: {
    disabled: false,
    config: {
      debug: true
    }
  }
```

DiederikvandenB:
Hi. See if this helps you out: https://github.com/nuxt-community/sentry-module/issues/19#issuecomment-384830231

eseceve:
After I update Nuxt to 2.43 and **@nuxt/sentry to 2.3.1** `$raven` nor `$sentry` aren't available on Store's actions.
How can I set an user in Sentry from my Store? I tried with this but doesn't works

```javascript
// store/user.js
export const actions = {
  async getProfile({ commit }, data) {
    const user = this.app.$axios.$get('/me/')
    this.app.$sentry.configureScope(scope => {
      scope.setUser(user)
    })
    // this.app.$sentry and this.$sentry are undefined
  }
}
```

drewbaker:
For future people finding this, this works for me with v4.0.3 as a Vuex action

```
    async GET_SELF({ commit }, email) {
        // Set user scope for Sentry
        this.app.$sentry.configureScope(function(scope) {
            scope.setUser({ email: email })
        })
        // Do your other stuff here (for me, that was Apollo request for user data)
    }
```

Webbist-dev:
> Also for future people finding this, 2020 was a wild year, I hope the future is better.

this aged like milk

Issue #19, opened 2018-03-22
I get undefined on my server log when accessing $raven from app context


pauldariye:
<img width="303" alt="screen shot 2018-03-22 at 02 38 54" src="https://user-images.githubusercontent.com/3650761/37755037-34386dee-2d7a-11e8-86d0-938d5e828701.png">

DiederikvandenB:
This is probably because you're trying to access $raven from the `asyncData` function, is that correct?

The following code works perfectly:

``` js
fetch({ $raven }) {
  console.log($raven);
}
```

This loads `raven-js`. If you want serverside reporting, just `import Raven from 'raven';`

DiederikvandenB:
I would if you can convince me why importing Raven as usual is not sufficient :). As far as I am aware (have not been using Nuxt/this module for a while now, just maintaining it), this approach will not result in your environment loading a new instance of Raven, and therefore I don't think it's necessary to inject it in to an SSR context.

Of course I can be wrong, and in that case I'm more than happy to help you out here!

pauldariye:
I get your point and I think I may have been calling it in the wrong place.

developius:
Does anyone have an example of using sentry to handle errors returned from `asyncData`?

Example page:

```js
export default {
  asyncData({ params, $api }) {
    // 404 thrown by backend service (uses axios underneath)
    return $api.getSomething(params.id).then(something => {
      // make `something` available to the page
      return { something }
    })
  }
}
```

```js
// plugins/api.js
import Vue from 'vue'
import ApiService from '@/services/ApiService'

export default (ctx, inject) => {
  const api = new ApiService({ $axios: ctx.app.$axios })
  ctx.$api = api
  inject('api', api)
}
```

```js
// services/ApiService.js
export default class ApiService {
  constructor(ctx) {
    this.$axios = ctx.$axios
  }

  getSomething(id){
    return this.$axios.get('/api/something/' + id)
  }
}
```

If a 404 is thrown by the API, I'd like nuxt to handle the error normally (which it does - it shows the `NuxtServerError` page) and an error to be logged in Sentry (not working 😢). Any ideas?

pauldariye:
@developius, you can wrap your api calls in a `try/catch` block or use `.catch` like so:

```javascript

export default {
  asyncData({ params, $api, app: { $raven } }) {
    // 404 thrown by backend service (uses axios underneath)
    return $api.getSomething(params.id).then(something => {
      // make `something` available to the page
[...]

PR #61, opened 2019-03-06
Add option to enable ssr

alexhyriavets:
Related to these issues: #19, #33 
Now it is not possible to log errors from **asyncData** or **nuxtServerInit** without some hacks as described in issues above.

Why we can't just add an option to enable/disable the plugin for server-side? Seems like it works just fine and we can (if `ssr: true`)
```javascript
asyncData(ctx) {
  try {
    const = await api.request()
  } catch (error) {
    ctx.app.$sentry.captureException(error)
  }
}
```

### DIFF
--- a/lib/sentry.js
+++ b/lib/sentry.js
@@ -9,6 +9,7 @@ module.exports = function sentry (moduleOptions) {
     dsn: process.env.SENTRY_DSN || false,
     disabled: process.env.SENTRY_DISABLED || false,
     disableClientSide: process.env.SENTRY_DISABLE_CLIENT_SIDE || false,
+    enableSsr: process.env.SENTRY_ENABLE_SSR || false,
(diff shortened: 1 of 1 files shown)


rchl:
The code that exposes $sentry object is exposing the `@sentry/browser` import. That doesn't seem like the right one to expose on the server side (should probably be `@sentry/node` there). So it doesn't seem like a correct fix to me.

rchl:
Also, a more thorough fix would be to:

 - change `sentry-client.js` file name as it would be no longer accurate
 - enable $sentry on ssr by default (don't see why it should be disabled by default)
 - the `disableClientSide` condition for adding the plugin should be changed as it should be possible to disable client side and still expose `$sentry` object on ssr.

aldarund:
Closing as invalid and implemented in https://github.com/nuxt-community/sentry-module/pull/72

Issue #68, opened 2019-04-02
Support sever side integration


pi0:
### What problem does this feature solve?

Node integration and allow using `$sentry` in server side code

Related issues: #19, #33
Related PRs: #61

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c52">#c52</a>)</em></sub></div>

rchl:
I think you meant `$sentry` rather than `$dsn`.
With latest changes dsn is set properly so sentry catches errors properly on the server side. Just that `$sentry` object is not exposed.

pi0:
Hi. Yes it was a after midnight typo 😂

aldarund:
closed in #72

Issue #63, opened 2019-03-08
About Client or Server logging


benoitemile:
### What problem does this feature solve?

We would be able to have specific monitoring and alerting on client side errors differently from server side errors.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c46">#c46</a>)</em></sub></div>

aldarund:
closed in #72

benoitemile:
thanks !!

PR #72, opened 2019-04-03
v3 update ( sentry 5, ssr plugin )

### LINKED ISSUE #70: [RFC] v3 checklist
- Server side plugin support
- Drop support for Nuxt 1.x
- Improve default webpack config
- Upgrade to sentry@5

aldarund:
sentry 5 update
plugin for server side
deprecate nuxt 1.0

Fixes #70 #68 #73 #63

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -82,12 +82,15 @@ module.exports = function sentry(moduleOptions) {
   // Register the server plugin
   if (!options.disableServerSide) {
     // Initialize Sentry
-    Sentry.init({
-      dsn: options.dsn,
-      ...options.config,
-      integrations: filterDisabledIntegration(options.serverIntegrations).
-        map(name => new Integrations[name](options.serverIntegrations[name]))
-    })
+    if (!options.config.disabled) {
+      Sentry.init({
+        dsn: options.dsn,
+        ...options.config,
+        integrations: filterDisabledIntegration(options.serverIntegrations).
+          map(name => new Integrations[name](options.serverIntegrations[name]))
+      })
+    }
+
     process.sentry = Sentry
     logger.success('Started logging errors to Sentry')
 
--- a/lib/module.js
+++ b/lib/module.js
@@ -46,6 +46,7 @@ module.exports = function sentry(moduleOptions) {
     this.addPlugin({
       src: path.resolve(__dirname, 'sentry.client.js'),
       fileName: 'sentry.client.js',
+      mode: 'client',
       options: {
         dsn: options.dsn,
         ...options.config
@@ -71,6 +72,7 @@ module.exports = function sentry(moduleOptions) {
     this.addPlugin({
       src: path.resolve(__dirname, 'sentry.server.js'),
       fileName: 'sentry.server.js',
+      mode: 'server',
       options: {
         dsn: options.dsn,
         ...options.config
--- a/lib/module.js
+++ b/lib/module.js
@@ -8,6 +8,7 @@ const logger = require('consola').withScope('nuxt:sentry')
 
 module.exports = function sentry(moduleOptions) {
   const publicPath = this.options.build.publicPath
+  const buildDirRelative = path.relative(this.options.rootDir, this.options.buildDir)
   const defaults = {
     dsn: process.env.SENTRY_DSN || false,
     disabled: process.env.SENTRY_DISABLED || false,
@@ -18,7 +19,7 @@ module.exports = function sentry(moduleOptions) {
       environment: this.options.dev ? 'development' : 'production'
     },
     webpackConfig: {
-      include: ['.nuxt/dist/client', '.nuxt/dist/server'],
+      include: [`${buildDirRelative}/dist`],
       ignore: [
         'node_modules',
         '.nuxt/dist/client/img'
--- a/lib/module.js
+++ b/lib/module.js
@@ -1,3 +1,4 @@
+import * as Integrations from '@sentry/integrations'
 const path = require('path')
 const Sentry = require('@sentry/node')
 const WebpackPlugin = require('@sentry/webpack-plugin')
@@ -6,21 +7,23 @@ const deepMerge = require('deepmerge')
 const logger = require('consola').withScope('nuxt:sentry')
 
 module.exports = function sentry(moduleOptions) {
+  const publicPath = this.options.build.publicPath
   const defaults = {
     dsn: process.env.SENTRY_DSN || false,
     disabled: process.env.SENTRY_DISABLED || false,
     disableClientSide: process.env.SENTRY_DISABLE_CLIENT_SIDE || false,
+    disableServerSide: process.env.SENTRY_DISABLE_SERVER_SIDE || false,
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     config: {
       environment: this.options.dev ? 'development' : 'production'
     },
     webpackConfig: {
-      include: ['.nuxt/dist/client'],
+      include: ['.nuxt/dist/client', '.nuxt/dist/server'],
       ignore: [
         'node_modules',
         '.nuxt/dist/client/img'
       ],
-      urlPrefix: '~/_nuxt/', // TODO: Use publicPath
+      urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
       configFile: '.sentryclirc'
     }
   }
@@ -38,12 +41,50 @@ module.exports = function sentry(moduleOptions) {
     return
   }
 
-  // Initialize Sentry
-  Sentry.init({
-    dsn: options.dsn,
-    ...options.config
-  })
-  logger.success('Started logging errors to Sentry')
+  // Register the client plugin
+  if (!options.disableClientSide) {
+    this.addPlugin({
+      src: path.resolve(__dirname, 'sentry.client.js'),
+      fileName: 'sentry.client.js',
+      options: {
+        dsn: options.dsn,
+        ...options.config
+      }
+    })
+  }
+
+  // Register the server plugin
+  if (!options.disableServerSide) {
+    // Initialize Sentry
+    Sentry.init({
+      dsn: options.dsn,
+      ...options.config,
+      integrations: [
+        new Integrations.Dedupe(),
+        new Integrations.ExtraErrorData(),
+        new Integrations.RewriteFrames(),
+        new Integrations.Transaction()
+      ]
+    })
+    logger.success('Started logging errors to Sentry')
+
+    this.addPlugin({
+      src: path.resolve(__dirname, 'sentry.server.js'),
+      fileName: 'sentry.server.js',
+      options: {
+        dsn: options.dsn,
+        ...options.config
+      }
+    })
+    this.nuxt.hook('render:setupMiddleware', app => app.use(Sentry.Handlers.requestHandler()))
+    this.nuxt.hook('render:errorMiddleware', app => app.use(Sentry.Handlers.errorHandler()))
+    this.nuxt.hook('generate:routeFailed', ({ route, errors }) => {
+      errors.forEach(({ error }) => Sentry.withScope((scope) => {
+        scope.setExtra('route', route)
+        Sentry.captureException(error)
+      }))
+    })
+  }
 
   // Enable publishing of sourcemaps
   this.extendBuild((config, { isClient, isDev }) => {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -39,7 +40,7 @@
     "standard-version": "^5.0.2"
   },
   "peerDependencies": {
-    "nuxt": "<1.0.0 || >1.2.1"
+    "nuxt": ">2.0.0"
```
pi0: As we have nuxt-edge do you think peerDependencies is needed?

lib/sentry.client.js
```diff
@@ -1,19 +1,15 @@
 import Vue from 'vue'
 import * as Sentry from '@sentry/browser'
+import * as Integrations from '@sentry/integrations';
```
pi0: Named import would help better treeshaking

lib/module.js
```diff
@@ -6,21 +7,23 @@ const deepMerge = require('deepmerge')
...
     disabled: process.env.SENTRY_DISABLED || false,
     disableClientSide: process.env.SENTRY_DISABLE_CLIENT_SIDE || false,
+    disableServerSide: process.env.SENTRY_DISABLE_SERVER_SIDE || false,
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     config: {
       environment: this.options.dev ? 'development' : 'production'
     },
     webpackConfig: {
-      include: ['.nuxt/dist/client'],
+      include: ['.nuxt/dist/client', '.nuxt/dist/server'],
```
pi0: ~> `.nuxt/dist` or maybe `{buildDir}/dist`?

lib/module.js
```diff
@@ -38,37 +41,41 @@ module.exports = function sentry(moduleOptions) {
...
+        new Integrations.Dedupe(),
+        new Integrations.ExtraErrorData(),
+        new Integrations.RewriteFrames(),
+        new Integrations.Transaction()
+      ]
+    })
+    logger.success('Started logging errors to Sentry')
+
+    this.addPlugin({
+      src: path.resolve(__dirname, 'sentry.server.js'),
```
pi0: Will the name here, imply only including in server bundle?

pi0: I suggest to specify `mode: 'server'` (Also this is not covering all 2.x support)

aldarund: @pi0 Yes, it should be processed by normalizePlugins as server only plugin, unless i miss something :)

aldarund: @pi0 hm, right . Is mode:server avaiable from 2.0 ? I think it was added later no?

pi0: Yes both added from a minor version. Using explicit mode would be better idea.

lib/module.js
aldarund: @pi0 is there a way to get buildDir relative path ? this.options.buildDir will have an absolute path and sentry-webpack-plugin wants here relative path

package.json
aldarund: ye, right, removed

lib/module.js
aldarund: I changed to use explicit mode, but since mode available only from some minor it means we will only support it from there. So need change readme. Or maybe u have better ideas?

lib/module.js
pi0: I would suggest something like this: `path.relative(rootDir, buildDir)`

lib/module.js
pi0: I think mentioning in the readme/changelog is sufficient. BW support with alias makes module code complex and unmaintainable.

lib/sentry.client.js
```diff
@@ -1,19 +1,15 @@
...
-      new Sentry.Integrations.ExtraErrorData,
-      new Sentry.Integrations.InboundFilters,
-      new Sentry.Integrations.FunctionToString,
-      new Sentry.Integrations.TryCatch,
-      new Sentry.Integrations.Breadcrumbs,
-      new Sentry.Integrations.GlobalHandlers,
-      new Sentry.Integrations.LinkedErrors,
-      new Sentry.Integrations.UserAgent,
-      new Sentry.Integrations.Vue({ Vue })
+      new Dedupe(),
```
pi0: Maybe we can make integrations configurable?

lib/sentry.server.js
```diff
@@ -0,0 +1,20 @@
...
+    integrations: [
+        new Dedupe(),
+        new ExtraErrorData(),
+        new RewriteFrames(),
+        new Transaction()
+      ]
+  })
+
+  if (!opts.disabled) {
+    Sentry.init(opts)
```
pi0: Is this init happening once or per-request? If sentry does not allows instance creation maybe it would be better to initialize it outside of this function and just inject it. (or move init to module and just inject here)

lib/module.js
```diff
@@ -38,37 +42,43 @@ module.exports = function sentry(moduleOptions) {
...
       }
     })
   }
 
-  // Initialize the hooks
-  if (this.nuxt.hook) {
+  // Register the server plugin
+  if (!options.disableServerSide) {
+    // Initialize Sentry
+    Sentry.init({
```
pi0: I think either this or `sentry.server.js` is duplicate as both are running in the same porcces. What do you think to Init sentry here and inject it to `process.sentry` so the server plugin can inject it directly as `$sentry`?

lib/module.js
```diff
@@ -2,25 +2,29 @@ const path = require('path')
 const Sentry = require('@sentry/node')
 const WebpackPlugin = require('@sentry/webpack-plugin')
 const deepMerge = require('deepmerge')
+const { Dedupe, ExtraErrorData, RewriteFrames, Transaction } = require('@sentry/integrations')
```
pi0: The node.js build does not allows any tree shaking. Maybe keeping original reference (`SentryIntegrations = `) would be better idea here so if we make integrations configurable we can use a for loop to enable some.

aldarund: @pi0 if we make integration configurable loop not enough, because integrations can take params, so we will need to pass whole initialized integration instances

aldarund: But then for client integrations we wont be able to serialize them ..

pi0: We can't do this for client config, maybe it can be in the form of an object that keys are integration name and values are an array that describes params? 

server:

```js
{
  Dedupe: [],
  ExtraErrorData: [],
  RewriteFrames: [],
  Transaction: []
}
```

client:

```js
{
  ExtraErrorData: [],
  ReportingObserver: [],
  RewriteFrames: []
}
```

- Regarding `VueIntegration`, it may have it's own config as we have to also pass Vue instance so need to somehow apply defaults
- Not sure if it is a convention or strict that params should be passed by a single object, but in that case, we may use values as `{}`

aldarund: @pi0 ye, that could work but params could be functions also.

pi0: We can partially support config or allow function serialization. In both cases supporting _what integration_ to enable is better than nothing :)

lib/sentry.server.js
```diff
@@ -0,0 +1,20 @@
...
+  const opts = Object.assign({}, <%= serialize(options) %>, {
+    integrations: [
+        new Dedupe(),
+        new ExtraErrorData(),
+        new RewriteFrames(),
+        new Transaction()
+      ]
+  })
+
+  if (!opts.disabled) {
```
rchl: Both in .server and .client, this check is no longer necessary because module will exit early if `disabled` is true.

lib/sentry.server.js
rchl: Exported plugin function is called per ssr request so this doesn't seem correct. And as you mentioned in other comment, Init is duplicated here and in the module for server.

lib/module.js
```diff
@@ -77,23 +87,14 @@ module.exports = function sentry(moduleOptions) {
...
-      renderer.errorMiddleware = (err, req, res, next) => {
-        // Log the error
-        Sentry.withScope((scope) => {
-          scope.setExtra('req', req)
-          Sentry.captureException(err)
-        })
+  // Enable publishing of sourcemaps
+  this.extendBuild((config, { isClient, isDev }) => {
+    if (!options.publishRelease || isDev) return
+    if (isClient) config.devtool = '#sourcemap'
```
rchl: Logic here was not, and still is not, correct IMO. This code here extends both client and server builds with the plugin and yet it only enables generation of source maps for client build. So this code will upload `dist` folder once on building client and then again on building server (without server source maps as those weren't generated).

So I think the proper logic here would be to either:
a) return early on `!isClient` so that we don't upload `dist/server` at all
or
b) always generate source maps and add  `WebpackPlugin` only for `!isClient` (as server bundles are generated second).

For me, I've actually decided to only upload client stuff here as I've enabled sourcemaps for the server in `nuxt.config.js` and I include those in deployed app and that seems to give proper stack traces already without uploading code to Sentry. It's possible that I'm missing out on something by not uploading to Sentry but I'm not sure what.
Given the above, I think I would appriciate a way to configure which source maps to upload to sentry (only client or both).

lib/module.js
```diff
@@ -2,25 +2,29 @@ const path = require('path')
...
-      include: ['.nuxt/dist/client'],
+      include: [`${buildDirRelative}/dist`],
       ignore: [
         'node_modules',
         '.nuxt/dist/client/img'
       ],
-      urlPrefix: '~/_nuxt/', // TODO: Use publicPath
+      urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
       configFile: '.sentryclirc'
     }
```
rchl: While at it, could you also set `release: options.config.release` on the webpack object?
That's because if one configures `release`, then both the one passed to `Sentry.init` and webpack config have to match. Otherwise Sentry will try to guess release name and I don't think it will be able to guess same thing on building and then running in production (as it needs git repo if it's guessing based on commits).

Reference: https://docs.sentry.io/platforms/node/sourcemaps/#uploading-source-maps-to-sentry

lib/module.js
rchl: I agree that Init is duplicated here and in `sentry.server.js`, but I I'm not sure what would be advantage of Initializing here and passing through `process.sentry`. Wouldn't it make sense to initialize in `sentry.server.js`, just not inside the exported function, but in global scope of the plugin, so that Sentry is not re-initialized on each request?

SarasArya:
Any idea when this will merge into master!! Waiting for this release eagerly. Btw great work you all.

lib/module.js
rchl: Now I understand... We need to pass options and can't initialize before that. So your solution sounds good to me.

lib/sentry.server.js
aldarund: @rchl its different check
disableClientSide vs config.disabled
 e.g. someone want to disable logging while leaving this.$sentry working in code as opposed to completely disabling. Before it was the same, so i just left logic as it is. Correct me if i wrong somewhere

lib/module.js
aldarund: added

lib/module.js
aldarund: Server maps generated by default as i see, so on second time it will upload both client and server. Also there a modern build so it can be three times :)

aldarund: @rchl i have added some logic to upload only once and allow disabling uploading server or client sources, although logic might be off

aldarund:
@rchl disabled flag not passed to sdk. Its just for not calling init as opposed to not include plugin at all. E.g. if u dont want your code crash locally with disabled sentry on error with this.$sentry is undefined

lib/module.js
```diff
@@ -2,31 +2,63 @@ const path = require('path')
...
   }
 
   const topLevelOptions = this.options.sentry || {}
   const options = deepMerge.all([defaults, topLevelOptions, moduleOptions])
+  if (!options.disableServerRelease) {
+    options.webpackConfig.include.push(`${buildDirRelative}/dist/server`)
```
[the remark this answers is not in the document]
rchl: Sorry, you were right in some other comment - server has source maps enabled by default [1].
Although I've never sent those to sentry and I seem to be getting proper stack traces there...

[1] https://github.com/nuxt/nuxt.js/blob/6ce99d2cc859c1cf17587acd4eccda468f160041/packages/webpack/src/config/server.js#L36

lib/module.js
```diff
@@ -38,37 +70,43 @@ module.exports = function sentry(moduleOptions) {
...
       }
     })
   }
 
-  // Initialize the hooks
-  if (this.nuxt.hook) {
+  // Register the server plugin
+  if (!options.disableServerSide) {
+    // Initialize Sentry
+    if (!options.config.disabled) {
```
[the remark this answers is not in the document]
aldarund: ye, thats valid points. It just was there before, but since its breaking release it fine to change it. 
I changed it to initialize option

lib/module.js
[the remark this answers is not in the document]
aldarund: Yes, that should be fine by default.

rchl:
LGTM

I would love to also have possibility to specify client only options as some are only supported on the client. Something like `configClient: {}` in default settings that we would `..options.configClient` to client options. We can handle that separately if you feel like there is too much feature creep here.

aldarund:
@rchl but if u pass not supported option it would be just ignored by sentry, right? If yes, then the usecase for client vs server configs would be only if same option should have different values on server vs client

rchl:
Not according to my tests. For example whitelistUrls setting would need different value for server and client. Documentation states that it's not supported on javascript client (only browser javascript), but in reality it's checked on both and on server stuff is ignored according to those rules, unexpectedly...

But on any case, for settings that take paths, it's not unrealistic to expect that client and server would need different values.

aldarund:
Ye, make sense. Then i guess config will be left, and then merged with clientConfig and serverConfig dicts

pi0:
You guys rock @aldarund @rchl  :heart:

meshuggie:
Any idea when you will be publishing this to NPM?

aldarund:
@meshuggie sorry for wait, its have been published now

Issue #67, opened 2019-04-02
add typescript defintions for $sentry


simllll:
### What problem does this feature solve?

right now $sentry is not available in the context or in a component. therefore no typescript definition is associated when using this.$sentry or in any other context.

### What does the proposed changes look like?

To solve this a proper typescript defintion is needed, right now to workaround this issue you simple can do:
1.) create a new "index.d.ts" file
2.) paste:

declare module 'vue/types/vue' {
    interface Vue {
        $sentry: any;
}
}
see also // https://vuejs.org/v2/guide/typescript.html#Augmenting-Types-for-Use-with-Plugins

but the correct solution would be not using "any", but the correct type. I tried it with:
import { NodeClient } from '@sentry/node';

but this collides with some other things, and actually it would be wrong on the client side too.

Could be the right way to provide a "unified" interface for both client&server and wrap the actual sentry calls behind it?

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c51">#c51</a>)</em></sub></div>

kevinmarrec:
Hi @simllll .

Indeed there is a need to define such typescript definitions so you get proper autocomplete & type checking. Declaring the type as `any` is also indeed the right way to go unless it gets properly defined.

Furthermore, you're again right about the fact both Server & Client seem to have different types.
Sentry instance in server comes from here : https://github.com/nuxt-community/sentry-module/blob/master/lib/module.js#L2, meanwhile Sentry instance in client comes from here : https://github.com/nuxt-community/sentry-module/blob/master/lib/sentry.client.js#L2.

Assuming `@sentry/node` source code https://github.com/getsentry/sentry-javascript/blob/master/packages/node/src/index.ts & how it gets implemented, it seems that on the server side of a Nuxt app, `$sentry` would be the whole package with all export.

It also seems that Sentry packages are individually exporting types : https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/index.ts#L1

@simllll Did you figure out maybe how can we grab the types ?

We need to first know how to get the types and how they're supposed to be used Server & Client sides, then we' should be able to define the type of `$sentry` properly.

rchl:
Client object is defined here https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/client.ts

I guess those types should work for both client and server?

simllll:
[...]

PR #84, opened 2019-05-15
feat: add typescript types for $sentry

simllll:
as discussed in #67 . Thanks @rchl for the hint 

regards

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,7 +1,22 @@
 import { Client } from '@sentry/types/dist/client';
 
+// add type to Vue context
 declare module "vue/types/vue" {
   interface Vue {
+    $sentry: Client;
+  }
+}
+
+// since nuxt 2.7.1 there is "NuxtAppOptions" for app context - see https://github.com/nuxt/nuxt.js/pull/5701
+declare module '@nuxt/vue-app' {
+  interface NuxtAppOptions {
+    $sentry: Client;
+  }
+}
+
+// add types for Vuex Store
+declare module 'vuex/types' {
+  interface Store<S> {
     $sentry: Client;
   }
 }
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-import { Client } from '@sentry/types/dist/client';
+import { Client } from '@sentry/types';
 
 // add type to Vue context
 declare module 'vue/types/vue' {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,0 +1,7 @@
+import { Client } from '@sentry/types/dist/client';
+
+declare module "vue/types/vue" {
+  interface Vue {
+    $sentry: Client;
+  }
+}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,7 +1,7 @@
 import { Client } from '@sentry/types/dist/client';
 
 // add type to Vue context
-declare module "vue/types/vue" {
+declare module 'vue/types/vue' {
   interface Vue {
     $sentry: Client;
   }
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,20 +3,24 @@ import { Client } from '@sentry/types';
 // add type to Vue context
 declare module 'vue/types/vue' {
   interface Vue {
-    $sentry: Client;
+    readonly $sentry?: Client;
   }
 }
 
-// since nuxt 2.7.1 there is "NuxtAppOptions" for app context - see https://github.com/nuxt/nuxt.js/pull/5701
-declare module '@nuxt/vue-app' {
+// App Context and NuxtAppOptions
+declare module '@nuxt/types' {
+  interface Context {
+    readonly $sentry?: Client;
+  }
+
   interface NuxtAppOptions {
-    $sentry: Client;
+    readonly $sentry?: Client;
   }
 }
 
 // add types for Vuex Store
 declare module 'vuex/types' {
   interface Store<S> {
-    $sentry: Client;
+    readonly $sentry?: Client;
   }
 }
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     }
   ],
   "files": [
-    "lib"
+    "lib",
+    "types"
   ],
   "main": "lib/module.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@sentry/integrations": "^5.2.0",
     "@sentry/node": "^5.2.0",
     "@sentry/webpack-plugin": "^1.6.2",
+    "@sentry/types": "^5.11.0",
     "consola": "^2.6.1",
     "deepmerge": "^3.2.0"
   },
--- a/package.json
+++ b/package.json
@@ -43,5 +43,6 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "types": "types/index.d.ts"
 }


rchl:
Now that Nuxt has [changed `context.app` type](https://github.com/nuxt/nuxt.js/pull/5701) from `Vue` to `NuxtAppOptions`, modules should start extending that one too. I don't know if that can be done in backwards compatible way though. Does importing and extending that type be a hard error if type doesn't exist. I guess it might? @kevinmarrec

simllll:
Interesting, but to be honest, unsure about it.
1.) Yes, you need the "base" file to extend from ,otherwise you get an error. 
2.) Even if this can be workaround, older nuxt version are not supported
3.) It doesn't give us any benefit of extending the NuxtAppOptions instead, I guess (haven't checked) NuxtAppOptions extends Vue. Otherweise a lot of other things would be probablay broken. 
To keep complexity low, I would prefet extending Vue, like all other plugins do already (and non-nuxt specific modules/plugins will do anyway).

Regards

rchl:
1, 2) If there is a way to extend types conditionally, based on availability of the type, then it would be backwards compatible.

> It doesn't give us any benefit of extending the NuxtAppOptions instead, I guess (haven't checked) NuxtAppOptions extends Vue

No, it doesn't extend Vue. It extends `ComponentOptions` which is just an object with key/value pairs, used to configure Vue instance.
That change didn't break things as `NuxtAppOptions` was currently created with `[key: string]: any` definition so it accepts anything. But as it had `Vue` type previously, modules got away with just extending `Vue` type while now that won't apply to `context.app` anymore.

rchl:
And to clarify, I'm not saying you should not extend `Vue`. I'm saying that modules should additionally start extending `NuxtAppOptions` with same key & type. Otherwise there will be no typescript help when using plugin object from `context.app.$sentry`.

kevinmarrec:
`context.app` had wrong type, it never have been a Vue instance here, it always have been ComponentOptions, it's just the naming (maybe `context.appOptions` would have been better name) and the old wrong type that have brought lot of foncusion.

Modules are supposed to extend types depending on what they're doing :
1 ) If the module uses `inject('sentry')` (to be accessible in all Vue instances as `this.$sentry`), the `Vue` interface need to be extended. (that's what is done in this PR for now)

2) If the module manually do `app.sentry` or whatever `app.anything`, then it needs to extend `NuxtAppOptions`

rchl:
@kevinmarrec `inject` handles injecting to all those places and more. It injects to `context.app` to `Vue` instance and even to `store`. So as long as plugin uses `inject` it ideally should extend interfaces of all those places.

kevinmarrec:
@rchl Alright then the following interfaces need to be extend with `$sentry` type :
`vue/types/vue` -> `Vue`
`@nuxt-vue-app` -> `NuxtAppOptions`
`vuex` -> `Store<S>`

rchl:
I agree, the question is whether it's possible to extend `NuxtAppOptions` in a way that would be compatible with Nuxt version lower than `2.7.1` which doesn't have that type.

kevinmarrec:
@rchl `ctx.app` had bad type before `2.7`, it was of type `Vue`.
Even if it was bad, it doesn't change nothing for modules which only `inject` like `sentry` one.

I mean, you'll be able to do `ctx.app.$sentry` in `<2.7.0` as we're extending `Vue` interface.

kevinmarrec:
But well I don't know if extending a non-existent interface can lead to problems, that's what you were probably asking

rchl:
> But well I don't know if extending a non-existent interface can lead to problems

Yes, that's what I'm wondering. If you try to import a type in `*.d.ts` file that doesn't exist it will probably break typescript but maybe it's possible to do try/catch (either literal or some equivalent). I don't have enough experience with TS to tell if it's possible.

kevinmarrec:
@rchl Well, as far as I know, if you do :

```ts
declare module 'mymodule' {
   interface MyInterface {
    myProperty: boolean
  }
}
```

If `MyInterface` exists, it extends it and if it doesn't it actually "create" it, and by "creating" I mean it provides this new interface as part of the module.

So IMO it shouldn't break anything, it will just provide useless interface for `<2.7`

rchl:
That sounds good to me then.

simllll:
@kevinmarrec Thanks, I just tested the same, I defintily have this problem in another project, but it seems as it is no issue in our nuxt project. Maybe related to some other typescript parameters.

```
declare module '@nuxt/vue-app-2' {
	// eslint-disable-next-line @typescript-eslint/interface-name-prefix
	interface NuxtAppOptionsNew {
		$a: void;
	}
}
```
doesn't break anything ;)

I've updated the PR.

Thanks @rchl for clarifying this. Learned something new today :).

Regards
SImon

types/index.d.ts
```diff
@@ -0,0 +1,22 @@
+import { Client } from '@sentry/types/dist/client';
+
+// add type to Vue context
+declare module "vue/types/vue" {
```
kevinmarrec:
```suggestion
declare module 'vue/types/vue' {
```

kevinmarrec:
@simllll Looks now god to me for client, what about server now ?

Assuming this line : https://github.com/nuxt-community/sentry-module/blob/master/lib/module.js#L113
We might need to extend `process` like here : https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/types/process.d.ts so `process.sentry` is typed.

rchl:
The common interface that Sentry exposes should work for both server and client. This pull request in fact uses it for both client and server. That's regarding your question: `If I remember there is a different sentry type depending if client or server, no ?`

As for `process.sentry`, I'm not sure the intention was to expose it publicly. It's more a necessity to be able to be able to access it and inject to `context` from a plugin, I believe. Right @aldarund ?
But maybe it would be useful in `process.sentry` too, to be used from `serverMiddleware`.

rchl:
> But maybe it would be useful in process.sentry too, to be used from serverMiddleware.

Actually `serverMiddleware`'s don't have access to those nuxt-specific `process` properties anyway so that argument is moot.

IMO there is no publicly expose `process.sentry` as it's already exposed everywhere where it's needed  using `inject` function.

kevinmarrec:
@rchl Alright, I think we're good then.

I'm not sure but I should able to simplify Nuxt modules type by doing this in core :

```ts
// @nuxt/vue-app/types/index.d.ts
export interface NuxtInject {
  [key: string]: any
}

export interface NuxtAppOptions extends ComponentOptions<Vue>, NuxtInject {}
```
```ts
// @nuxt/vue-app/types/vue.d.ts
declare module 'vue/types/vue' {
  interface Vue extends NuxtInject {
    $nuxt: NuxtApp
  }
}
```
```ts
// @nuxt/vue-app/types/vuex.d.ts
declare module 'vuex/types' {
  interface Store<S> extends NuxtInject {
    $nuxt: NuxtApp
  }
}
```

And then modules which use `inject` will just have to provide one simple type, example for sentry module :

```ts
import { Client } from '@sentry/types/dist/client'

declare module '@nuxt/vue-app/types' {
  interface NuxtInject {
    $sentry: Client
  }
}
```

I tried and the `extends NuxtInject` seems to work.

But for now let's stick with this PR as everything is good :)

simllll:
I'm currently working on this, as it turned out the sentry/node and sentry/browser do NOT have the same interface. I'm looking into options if we can abstract it somehow that the "sentry/client" interface (which is the one for browser) can be used also on server side.

I'll update this PR as soon as I have an valid approach.

Regards
Simon

Atinux:
Friendly ping @simllll @kevinmarrec about this <3

simllll:
Hi @Atinux , sorry for forgetting about this thread, we have solved it like it is in this PR already.
Interface for client and server is the same (in the meantime?). Just double checked it.

e.g. this is this.$sentry.captureException on SSR and on Client:
```js
function captureException(exception) {                                                                                                                                                                                                        var syntheticException;
    try {
        throw new Error('Sentry syntheticException');
    }
    catch (exception) {
        syntheticException = exception;
    }
    return callOnHub('captureException', exception, {
        originalException: exception,
        syntheticException: syntheticException,
    });
}
```
We have a type defintion of this in our app code to cover it:
```

declare module 'vue/types/vue' {
	// eslint-disable-next-line @typescript-eslint/interface-name-prefix
	interface Vue {
...
		$sentry: Client;
...
	}
}
```

kevinmarrec:
It's now `@nuxt/types` over `@nuxt/vue-app` (since **Nuxt 2.8** AFAIK), so it needs to be updated.
Also my last comment hasn't been taken into account, `types` folder need to be specified in `files` entry of `package.json` to be correctly uploaded to npm later on.

rchl:
Also per #2158 the interfaces are not the same on server and client so there is no ideal way to fix that, I think.
It might be OK to use client types on both, for example -- it might cover the typical use, but it won't be 100% correct.

simllll:
> It's now `@nuxt/types` over `@nuxt/vue-app` (since **Nuxt 2.8** AFAIK), so it needs to be updated.
> Also my last comment hasn't been taken into account, `types` folder need to be specified in `files` entry of `package.json` to be correctly uploaded to npm later on.

I just updated to @nuxt/types, added also the context, and made $sentry optional due to fact that the disabled flag deactivates the injectino of $sentry at all.

Reagrding package.json I cannot follow, it's in there already?
image

Thanks

simllll:
> Also per #2158 the interfaces are not the same on server and client so there is no ideal way to fix that, I think.
> It might be OK to use client types on both, for example -- it might cover the typical use, but it won't be 100% correct.

Both implementations are using @sentry/core behind the scene:
@sentry/node/dist/index.d.ts 
`export { addGlobalEventProcessor, addBreadcrumb, captureException, captureEvent, captureMessage, configureScope, getHubFromCarrier, getCurrentHub, Hub, Scope, setContext, setExtra, setExtras, setTag, setTags, setUser, withScope, } from '@sentry/core';
`
@sentry/browser/dist/export.d.ts:
`export { addGlobalEventProcessor, addBreadcrumb, captureException, captureEvent, captureMessage, configureScope, getHubFromCarrier, getCurrentHub, Hub, Scope, setContext, setExtra, setExtras, setTag, setTags, setUser, withScope, } from '@sentry/core';`

So should be identical in my opinion?

rchl:
> Both implementations are using @sentry/core behind the scene:

I was basing my comment on what Sentry author wrote. If so, then it looks fine and I'm not sure why Sentry author saw it being hard to fix.

types/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+declare module 'vue/types/vue' {
+  interface Vue {
+    $sentry?: Client;
+  }
+}
+
+// App Context and NuxtAppOptions
+declare module '@nuxt/types' {
+  interface Context {
+    readonly $snack?: Client;
```
rchl: `$snack`?

kevinmarrec: 🙃

simllll: uha ;) sorry about that, fixed it. 
i rebased and squashed all the commits, unfortuantely I guess due to missing write permissions this is not reflected in this pr.

types/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+declare module 'vue/types/vue' {
+  interface Vue {
+    readonly $sentry?: Client;
+  }
+}
+
+// App Context and NuxtAppOptions
+declare module '@nuxt/types' {
+  interface Context {
+    readonly $sentry?: Client;
```
rchl: Having `$sentry` optional, while more correct, will make using it quite inconvenient since you'll have to null check it always.

It sucks a bit...

simllll: you are right, but I can tell that we ran into this already several times as we have turned off sentry on development..
Thanks to optional chaining this should be easily doable ;) just do this.$sentry?.captureException() (not tested, but guess this should be fine)

kevinmarrec: Hm, sounds like a good catch, if you're saying `@nuxtjs/sentry` in types of `tsconfig.json` it means you're gonna use it and it should be defined, so it's indeed inconvenient having it optional here

rchl: But instead of disabling you could set `initialize: false` instead.

Although if this module allows disabling (maybe it shouldn't), it wouldn't be right to advocate on making it non-optional. Unless we remove option to "disable".

rchl: And BTW. I can't use optional chaining so I haven't switched to TS yet. For now using JS + type checking.

simllll: well, if it would be there all the time, I would fully agree, but it is runtime dependent if it is there or not (regarding to disabled). Another solution would be to inject a $sentry "dummy" on disabled state, that just outputs a warning on all function calls that sentry is disabled and therfore not logging anything. Even though I really would like this approach, I would say this is another story / PR.

kevinmarrec: I'm not definitely sure but I think **not having** the type optional is better

rchl: @kevinmarrec even though I don't like it myself, I think we shouldn't make it non-optional unless the module's code is changed to make it always exist. Otherwise, we are potentially lying with types and that can trip people and cause bugs.

kevinmarrec: Sure, so we're good with current status right ?

kevinmarrec:
Btw: No idea what's going on with codecov checks :D

types/index.d.ts
rchl: I'm not happy but as far as this PR it LGTM.

Atinux: Merging now, can be improved later on :)

types/index.d.ts
simllll: checkout #136 for a possible solution to this ;)

types/index.d.ts
simllll: please check out #136 for a possible solution :)

rchl:
@simllll The `Client` interface seems to be missing quite a few methods per what is here:
https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/client.ts

In https://github.com/nuxt-community/sentry-module/pull/84#issuecomment-579774335 you've posted an export that has more methods (for example `setExtra`, `setExtras`, ...) but it's not the one that is actually used here for `$sentry`.

simllll:
> @simllll The `Client` interface seems to be missing quite a few methods per what is here:
> https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/client.ts
> 
> In [#84 (comment)](https://github.com/nuxt-community/sentry-module/pull/84#issuecomment-579774335) you've posted an export that has more methods (for example `setExtra`, `setExtras`, ...) but it's not the one that is actually used here for `$sentry`.

This has been changed in sentry since they have the new JS sdk.

You only can use the Client exported functions ..e.g. captureException, captureMessage etc..
if you wanna append extra data (e.g. setExtra) you have to do it like this:
```js
this.$sentry.withScope(scope => {
						scope.setExtra('address', address);
						scope.setLevel('level');
						this.$sentry.captureException(err);
					});

```
it was like this already before in the old sdk for the server, but not for the client one. But now they have changed this everywhere to this style. There is some explanation why they did it, but couldn't find it right now.

rchl:
Thanks for the explanation.
Seems like I was still using old methods in my code but maybe they kept backwards compatibility for now as I think that code was still working.

rchl:
Although `withScope` is also not in those types...

simllll:
Indeed, weird that this didn't show up before on my side.. but I clearly can see that withScope is not in there, 
import * as SentryTypes from '@sentry/core';

$sentry: typeof SentryTypes;

would work.. I do some more checks to be safe and compare client and server side again

PR #85, opened 2019-05-28
docs: provide informations about server side usage

felixdenoix:
HUGE THANKS FOR THIS MODULE !
I had some issues trying to find out how to use the sentry module server side as I wasn't sure of were the `$sentry` module was injected so I thought this paragraph might be useful.

PR #94, opened 2019-08-01
add info about usage from asyncData

rchl:
Add info about usage from asyncData and store and remove inaccurate
comments about injection (it's inject to context.app and not context).


rchl:
Hi @aldarund , please have a look

rchl:
Actually I forgot that there is #85 already which is more detailed.

PR #123, opened 2019-11-05
Set release to auto by default

### LINKED ISSUE #121: Attaching commits to release not working
### Version

[3.0.1](https://github.com/sentry-module/releases/tag/3.0.1)

### Reproduction link

[https://jsfiddle.net](https://jsfiddle.net)

### Steps to reproduce

1. Make sure you are in an active git repo
2. Install this package (`yarn add @nuxtjs/sentry`)
3. In your `nuxt.config.js`, add the following:

```js
export default {
    // ...

    modules: [
        // ...
        '@nuxtjs/sentry',
    ],

    sentry: {
        dsn: 'https://your-dsn@sentry.io/project-id',
        publishRelease: true
    },

    // ...
}
```

4. run `yarn build`
5. The script will create a release with Sentry using the current git commit hash, upload sourceMaps, set the release in the sentry client but not associate git commits to the release.

### What is expected ?

It should also attach commits to the release (with the `--auto` flag)

It is not attaching any commits

### Additional comments?

I have already looked at #65 but that didn't resolve my issue.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c74">#c74</a>)</em></sub></div>

georgeboot:
Fixes #121 

It sets the `--auto` flag to `true`. This means that when `publishRelease` is `true`, it will by default auto discover the repo, last commit etc.

If users don't like this behaviour and they can one of two things:

1. Disable attaching commits completely by setting env`SENTRY_AUTO_ATTACH_COMMITS=false` or `attachCommits: false` in config.
2. Set the commit themselves in the `nuxt.config.js`:

```js
export default {
  // ...
  sentry: {
        dsn: 'https://your-dsn@sentry.io/project-id',
        webpackConfig: {
            setCommits: {
                repo: 'organisation/repo-name',
                commit: 'your-commit-hash',
                auto: false,
            },
        },
    },
}
```

Before this should be merged, two PRs at Sentry should be merged first:

- [x] https://github.com/getsentry/sentry-webpack-plugin/pull/156
- [x] https://github.com/getsentry/sentry-cli/pull/618

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -21,7 +21,7 @@ module.exports = function sentry (moduleOptions) {
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     disableServerRelease: process.env.SENTRY_DISABLE_SERVER_RELEASE || false,
     disableClientRelease: process.env.SENTRY_DISABLE_CLIENT_RELEASE || false,
-    attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS === 'true' || true,
+    attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS !== '0',
     repo: process.env.SENTRY_RELEASE_REPO || false,
     clientIntegrations: {
       Dedupe: {},
--- a/lib/module.js
+++ b/lib/module.js
@@ -48,7 +48,7 @@ module.exports = function sentry (moduleOptions) {
       urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
       configFile: '.sentryclirc',
       release: {
-        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || true,
+        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || false
       }
     }
   }
--- a/lib/module.js
+++ b/lib/module.js
@@ -21,6 +21,8 @@ module.exports = function sentry (moduleOptions) {
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     disableServerRelease: process.env.SENTRY_DISABLE_SERVER_RELEASE || false,
     disableClientRelease: process.env.SENTRY_DISABLE_CLIENT_RELEASE || false,
+    attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS || true,
+    repo: process.env.SENTRY_RELEASE_REPO || false,
     clientIntegrations: {
       Dedupe: {},
       ExtraErrorData: {},
@@ -46,10 +48,7 @@ module.exports = function sentry (moduleOptions) {
         '.nuxt/dist/client/img'
       ],
       urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
-      configFile: '.sentryclirc',
-      release: {
-        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || true
-      }
+      configFile: '.sentryclirc'
     }
   }
 
@@ -68,6 +67,16 @@ module.exports = function sentry (moduleOptions) {
 
   if (options.config.release && !options.webpackConfig.release) {
     options.webpackConfig.release = options.config.release
+  }
+
+  if (options.attachCommits) {
+    options.webpackConfig.setCommits = {
+      auto: true
+    }
+
+    if (options.repo) {
+      options.webpackConfig.setCommits.repo = options.repo
+    }
   }
 
   const initializationRequired = options.initialize && options.dsn
--- a/lib/module.js
+++ b/lib/module.js
@@ -46,7 +46,10 @@ module.exports = function sentry (moduleOptions) {
         '.nuxt/dist/client/img'
       ],
       urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
-      configFile: '.sentryclirc'
+      configFile: '.sentryclirc',
+      release: {
+        auto: true,
+      },
     }
   }
 
--- a/lib/module.js
+++ b/lib/module.js
@@ -48,7 +48,7 @@ module.exports = function sentry (moduleOptions) {
       urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
       configFile: '.sentryclirc',
       release: {
-        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || false
+        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || true
       }
     }
   }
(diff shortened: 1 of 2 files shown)


georgeboot:
@aldarund can you have a look please?

georgeboot:
@pi0 ping

lib/module.js
```diff
@@ -46,7 +46,10 @@ module.exports = function sentry (moduleOptions) {
         '.nuxt/dist/client/img'
       ],
       urlPrefix: publicPath.startsWith('/') ? `~${publicPath}` : publicPath,
-      configFile: '.sentryclirc'
+      configFile: '.sentryclirc',
+      release: {
+        auto: process.env.SENTRY_AUTO_ATTACH_COMMITS || true
```
rchl: Aren't env vars always strings? Thus `process.env.SENTRY_AUTO_ATTACH_COMMITS`, when set to `false` will have a `"false"` string and will also evaluate to `true`.

rchl:
> If users don't like this behaviour and they can one of two things:

I  think that information should be documented around `publishRelease` option documentation.

lib/module.js
georgeboot: I stole it from here: https://github.com/nuxt-community/sentry-module/blob/master/lib/module.js#L21

georgeboot: It's actually being used like that evereywhere:
https://github.com/nuxt-community/sentry-module/blob/984bae6a41f99395221d7cf4cc0d1f04686ac8c5/lib/module.js#L15-L23

rchl: Yes, but since those default to `false` (apart from `SENTRY_INITIALIZE` which will probably have the same issue), those will actually work by accident - setting ENV to something will make it evaluate to truthy value thus overriding `false`.

Even if it would not be a string (which it is), you wouldn't be able to override it with `false` as `false || true` still evaluates to `true`.

georgeboot:
Updated the working a bit (now uses config options) and added docs.

lib/module.js
georgeboot: True. Fixed

lib/module.js
```diff
@@ -21,6 +21,8 @@ module.exports = function sentry (moduleOptions) {
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     disableServerRelease: process.env.SENTRY_DISABLE_SERVER_RELEASE || false,
     disableClientRelease: process.env.SENTRY_DISABLE_CLIENT_RELEASE || false,
+    attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS === 'true' || true,
```
rchl: It's still not possible to override to `false`. `false || true` is still `true`.

I think it would make sense to have some helper function for handling that. Like:
```js
function getEnvironmentValueOrDefault(envKey, defaultValue) {
  if (process.env[envKey] !== undefined) {
    return process.env[envKey] === 'true'
  }

  return defaultValue
}
// ...
attachCommits: getEnvironmentValueOrDefault('SENTRY_AUTO_ATTACH_COMMITS', true),
```

rchl: Even better would be to maybe treat `1` or `true` as truthy values and `0` or `false` as falsey. I think people often use 0/1 for env variables.

georgeboot: Oh wow completely missed that. My mind was telling me that `||` is JavaScripts nullable operator but obviously it isn’t.

georgeboot: Fixed now, I guess

lib/module.js
```diff
@@ -21,6 +21,8 @@ module.exports = function sentry (moduleOptions) {
     publishRelease: process.env.SENTRY_PUBLISH_RELEASE || false,
     disableServerRelease: process.env.SENTRY_DISABLE_SERVER_RELEASE || false,
     disableClientRelease: process.env.SENTRY_DISABLE_CLIENT_RELEASE || false,
+    attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS !== '0',
```
rchl: Sorry if I wasn't clear but I think `false` also needs to be treated as falsey value. I meant to handle 0 as an extra.
Still would suggest extra helper function.

rchl: Still not addressed but was resolved.

aldarund: why zero? 
Better to be same as other option e.g.
process.env.SENTRY_AUTO_ATTACH_COMMITS || false

alexcbz:
I'll help a lot to have auto commits. Thanks @georgeboot!

rchl:
So did we end up with this:
```
attachCommits: process.env.SENTRY_AUTO_ATTACH_COMMITS !== '0',
```
?

Does that mean that `attachCommits` ends up being enabled by default? That sounds like a possibly breaking change to me.

P4sca1:
This just broke my release pipeline.
With `auto: false` in `sentry.webpackConfig.setCommits`, the `--auto` flag is still added.
@georgeboot

rchl:
@P4sca1 create PR #138 to disable it by default but I think you are supposed set `attachCommits` to false to disable, not change `auto` to `false`, which you would need to have to change the source code to do.

P4sca1:
@rchl I don’t want to disable attaching commits completely, because I manually assign a commit. IMO the auto setting should regulate whether the auto flag is set.
Also I agree that changing the auto flag default value in a non-breaking (non-major) release feels wrong.

rchl:
It feels like that feature was added prematurely, without proper testing or documentation.

But since you can just pass `webpackConfig: { ... }` options directly to the module (sort of undocumented feature), you can set `attachCommits` to false and then manually set the options you want.

P4sca1:
Where is this attachCommits config coming from?
Because I can’t find it here: https://github.com/getsentry/sentry-webpack-plugin/blob/master/README.md

rchl:
@P4sca1 `attachCommits` is a configuration name specific to this module. It set the `setCommits` object on the configuration passed to `sentry-webpack-plugin`. So you can pass your own configuration using something like:
```js
['sentry-module', {
  dsn: '111',
  // ...
  webpackConfig: {
    setCommits: {
      // ...
    }
  }
}]

**Comment by P4sca1** [comment]
Also why does auto: false no longer works when it should pass it directly to the sentry webpack plugin?

**Comment by rchl** [comment]
You could check code of `sentry-webpack-plugin`, I'm not using commits release so not sure.
Are you sure that you have actually changed the code to pass `auto: false`? For that you would need to change the source code of this module.

**Comment by P4sca1** [comment]
In the first post, it says that you can set `auto: false` to disable the --auto flag.

**Comment by weotch** [comment]
This is what ended up working for me when deploying via Netlify and using GitLab for repo:

```js
  sentry: {
    publishRelease: true,
    attachCommits: false,
    config: {
      release: process.env.COMMIT_REF, // From Netlify
      environment: 'production',
      extra: { // Netlify env variables
        url: process.env.URL,
        deploy_url: process.env.DEPLOY_URL // From Netlify
      }
    },
    webpackConfig: {
      setCommits: {
        commit: process.env.COMMIT_REF, // From Netlify
        repo: 'Organization / SubGroup / ProjectName'  // Using GitLab with subgroups
      }
    }
  }
```

PR #136, opened 2020-01-29
feat: mock sentry in case it is disabled

simllll:
as discussed in #84 this would be a way to ensure $sentry is there all the time.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -81,17 +81,12 @@ module.exports = function sentry (moduleOptions) {
 
   const initializationRequired = options.initialize && options.dsn
 
-  if (options.disabled) {
-    logger.info('Errors will not be logged because the disable option has been set')
-    return
-  }
-
   if (!options.dsn) {
     logger.info('Errors will not be logged because no DSN has been provided')
   }
 
   // Register the client plugin
-  if (!options.disableClientSide) {
+  if (!options.disabled && !options.disableClientSide) {
     this.addPlugin({
       src: path.resolve(__dirname, 'sentry.client.js'),
       fileName: 'sentry.client.js',
@@ -109,10 +104,17 @@ module.exports = function sentry (moduleOptions) {
           }, {})
       }
     })
+  } else {
+    logger.info('Sentry client side errors will not be logged because the disable option has been set')
+    this.addPlugin({
+      src: path.resolve(__dirname, 'sentry.mocked.js'),
+      fileName: 'sentry.client.js',
+      mode: 'client'
+    })
   }
 
   // Register the server plugin
-  if (!options.disableServerSide) {
+  if (!options.disabled && !options.disableServerSide) {
     // Initialize Sentry
     if (initializationRequired) {
       Sentry.init({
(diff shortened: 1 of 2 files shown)


rchl:
```
/Users/../sentry-module/lib/module.js
  12:33  error  Missing space before function parentheses  space-before-function-paren

/Users/../sentry-module/lib/sentry.mocked.js
  3:5   error    Expected method shorthand                                                                                      object-shorthand
  5:9   warning  Unexpected console statement                                                                                   no-console
  5:95  error    Closing curly brace should be on the same line as opening curly brace or on the line after the previous block  brace-style
```

simllll:
> ```
> /Users/../sentry-module/lib/module.js
>   12:33  error  Missing space before function parentheses  space-before-function-paren
> 
> /Users/../sentry-module/lib/sentry.mocked.js
>   3:5   error    Expected method shorthand                                                                                      object-shorthand
>   5:9   warning  Unexpected console statement                                                                                   no-console
>   5:95  error    Closing curly brace should be on the same line as opening curly brace or on the line after the previous block  brace-style
> ```

Thanks, I get following eslint errors:
```
ESLint: 6.8.0.

ESLint couldn't find the plugin "eslint-plugin-vue".

```
after I add this package, I get other weird errors..e.g.
```
Error: .eslintrc » @nuxtjs/eslint-config:
        Environment key "jest/globals" is unknown

```

rchl:
The use of `Proxy` is neat but I believe it's not transpiled so such code will fail on IE11 (https://caniuse.com/#search=proxy).

Is it OK? Maybe it is because I imagine in production people would not have Sentry disabled. But who knowns.

simllll:
> The use of `Proxy` is neat but I believe it's not transpiled so such code will fail on IE11 (https://caniuse.com/#search=proxy).
> 
> Is it OK? Maybe it is because I imagine in production people would not have Sentry disabled. But who knowns.

Same thoughts, but adding sentry plugin and disabled it wouldn't make any sense to me at all. Therefore I would vote for ignoring this evil browser for this case ;). (Maybe add a hint in the docs though)

lib/module.js
```diff
@@ -81,17 +81,12 @@ module.exports = function sentry (moduleOptions) {
 
   const initializationRequired = options.initialize && options.dsn
 
-  if (options.disabled) {
-    logger.info('Errors will not be logged because the disable option has been set')
-    return
```
rchl: Maybe keep the log, just without return.

lib/module.js
```diff
@@ -139,26 +141,35 @@ module.exports = function sentry (moduleOptions) {
         Sentry.captureException(error)
       }))
     })
+  } else {
+    logger.info('Sentry server side errors will not be logged because the disable option has been set')
+    this.addPlugin({
+      src: path.resolve(__dirname, 'sentry.mocked.js'),
+      fileName: 'sentry.mocked.js',
```
rchl: I think you can skip filename to have it autogenerated.

This way you are also overriding the same file twice, if both server and client are disabled. That can be a problem since code can get transpiled differently for each.

rchl: Maybe the argument about overwriting the file is not true but anyways there is no need to specify file name.

simllll: Ah, thanks. I will set it to sentry.server.js and sentry.client.js?

lib/module.js
simllll: I added     logger.info('Sentry client side errors will not be logged because the disable option has been set')
and     logger.info('Sentry server side errors will not be logged because the disable option has been set')
instead further down the code :)

lib/module.js
simllll: > Maybe the argument about overwriting the file is not true but anyways there is no need to specify file name.

I changed it to server and client, as this was already the case for the regular if.

rchl:
This should have also updated types. I'll create new PR.

simllll:
> This should have also updated types. I'll create new PR.

Indeed, somehow this got lost.. I had the type change already in it ..

PR #138, opened 2020-01-30
Disable commit releasing by default since it's a breaking change

rchl:
That change shouldn't have been released in minor version as it
is likely to break workflows since releasing commit often needs extra
configuration/work on the repo and/or sentry side.

For example for me it caused:
[error] Command failed: /var/lib/jenkins/jobs/workspace/node_modules/@sentry/cli/sentry-cli releases set-commits 2.22.107 --auto
error: API request failed
  caused by: sentry reported an error: You do not have permission to perform this action. (http status: 403)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Normally setting required DSN information would be enough.
 
 ### attachCommits
 - Type: `Boolean`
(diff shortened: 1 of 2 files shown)


Atinux:
Thank you @rchl, I let you keep the lead on this module 👍

PR #147, opened 2020-02-12
DX - Adjust log message

robsontenorio:
A hint where the log message comes from.

message


rchl:
I think the elephant in the room (as your screenshot illustrates) is that there are two messages, one saying that Sentry will not log errors due to DSN being missing and another one saying that Sentry started logging errors.

As for the prefix that you've added, it's easy to miss but there is already such "tag" shown on the far right of the line. So it's a bit redundant to add it at the beginning. And if we really want it, probably we should add it to all messages shown by this module.

PR #153, opened 2020-02-12
fix(types): add types for process.sentry

rchl:
Extending a namespace must be done in a definition file that has no
top-level imports so that the file is treated as ambient module
rather than a normal module. For that reason, I couldn't just put new
code in `index.d.ts` but had to create separate file. Stuff from
`index.d.ts` moved to `extend.d.ts` to have a cleaner structure.


rchl:
@simllll feel free to have a look/review

PR #157, opened 2020-02-24
Don't override an existing webpack config "devtool" setting

alexblunck:
This allows using a custom `devtool` setting, i.e. `hidden-source-map`.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -168,7 +168,7 @@ export default function SentryModule (moduleOptions) {
         return
       }
 
-      config.devtool = 'source-map'
+      config.devtool = config.devtool || 'source-map'
 
       // when not in spa mode upload only at server build
       if (isClient && this.options.mode !== 'spa') {


rchl:
Nuxt sets default values of `devtool`.

In dev: cheap-module-eval-source-map (client), cheap-module-source-map (server)
When building: false (client), cheap-module-source-map (server)

With your change, this module will be unable to set `source-map` value in all those cases where there is already default value set. You could argue that in dev it doesn't matter and in fact sentry shouldn't even run in dev.
But when building for deployment, it would be sort of unreliable to rely on what Nuxt sets by default so I'm not sure that change is good.

Could you explain the motivation for your change?
Does using `hidden-source-map` generally make sense for Sentry? Is Sentry able to map code in errors (especially on the client) when using that value?
Does it make sense to just change to that value?

rchl:
I've been testing a little bit and I think `hidden-source-map` is a good default value since Sentry will still be able to lookup source from minified stack.

Bonus is that the browser will not complain about not finding source maps when opening developer tools on a production site.

If we want to allow customizing devtool setting, we can do that through module's setting instead but for now, I think it's enough to just change the defaults.

The only concern is whether that will be a breaking change or not. It's not directly breaking Sentry but maybe some will also want to use those generated files in a different way and that workflow might be affected.

mstaack:
> Bonus is that the browser will not complain about not finding source maps when opening developer tools on a production site.

has been one reason for me ;)

rchl:
Actually I'm a bit afraid of introducing that change in a non-major release so added possibility to configure it instead - #158

alexblunck:
Being able to configure it sounds great!

PR #167, opened 2020-03-12
Setup WebpackPlugin during webpack:config hook

matthieusieben:
If a custom webpack config is added by another module, that other config would **not** have the `WebpackPlugin` added to its plugins since `extendBuild` is called only when NuxtJS build the initial list of `webpackConfigs`, and not on the actual list of webpack configurations. The `webpack:config` can be used to manipulate the actual list of webpack configuration that will be used, even those added by a custom module.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -164,24 +164,30 @@ export default function SentryModule (moduleOptions) {
 
   // Enable publishing of sourcemaps
   if (!options.disabled) {
-    this.extendBuild((config, { isClient, isModern, isDev }) => {
-      if (!options.publishRelease || isDev) {
+    this.nuxt.hook('webpack:config', async webpackConfigs => {
+      if (!options.publishRelease || this.options.dev) {
         return
       }
 
-      config.devtool = options.sourceMapStyle
+      webpackConfigs.forEach(config => {
+        config.devtool = options.sourceMapStyle
 
-      // when not in spa mode upload only at server build
-      if (isClient && this.options.mode !== 'spa') {
-        return
-      }
-      // when in spa mode upload only at modern build if enabled
-      if (isClient && !isModern && this.options.modern) {
-        return
-      }
+        const isClient = config.name !== 'server'
+        const isModern = config.name === 'modern'
 
-      config.plugins.push(new WebpackPlugin(options.webpackConfig))
-      logger.info('Enabling uploading of release sourcemaps to Sentry')
+        // when not in spa/universal mode upload only at server build
+        if (isClient && this.options.mode !== 'spa' && this.options.mode !== 'universal') {
+          return
+        }
+        // when in spa/universal mode upload only at modern build if enabled
+        if (isClient && !isModern && this.options.modern) {
+          return
+        }
+
+        config.plugins = config.plugins || []
+        config.plugins.push(new WebpackPlugin(options.webpackConfig))
+        logger.info(`Enabling uploading of release sourcemaps to Sentry for ${config.name}`)
+      })
     })
   }
 }
--- a/lib/module.js
+++ b/lib/module.js
@@ -172,7 +172,7 @@ export default function SentryModule (moduleOptions) {
       webpackConfigs.forEach((config) => {
         config.devtool = options.sourceMapStyle
 
-        const isClient = config.name !== 'server'
+        const isClient = config.name === 'modern' || config.name === 'client'
         const isModern = config.name === 'modern'
 
         // when not in spa/universal mode upload only at server build
--- a/lib/module.js
+++ b/lib/module.js
@@ -164,7 +164,7 @@ export default function SentryModule (moduleOptions) {
 
   // Enable publishing of sourcemaps
   if (!options.disabled) {
-    this.nuxt.hook('webpack:config', async webpackConfigs => {
+    this.nuxt.hook('webpack:config', (webpackConfigs) => {
       if (!options.publishRelease || this.options.dev) {
         return
       }
--- a/lib/module.js
+++ b/lib/module.js
@@ -164,27 +164,23 @@ export default function SentryModule (moduleOptions) {
 
   // Enable publishing of sourcemaps
   if (options.publishRelease && !options.disabled && !this.options.dev) {
+    if (this.options.build.parallel) {
+      throw new Error('@nuxtjs/sentry is not able to publish a release in parallel build mode.')
+    }
+
     this.nuxt.hook('webpack:config', (webpackConfigs) => {
       webpackConfigs.forEach((config) => {
         config.devtool = options.sourceMapStyle
-
-        const isClient = config.name === 'modern' || config.name === 'client'
-        const isModern = config.name === 'modern'
-
-        // when not in spa/universal mode upload only at server build
-        if (isClient && this.options.mode !== 'spa' && this.options.mode !== 'universal') {
-          return
-        }
-        // when in spa/universal mode upload only at modern build if enabled
-        if (isClient && !isModern && this.options.modern) {
-          return
-        }
-
-        config.plugins = config.plugins || []
-        config.plugins.push(new WebpackPlugin(options.webpackConfig))
-
-        logger.info(`Enabling uploading of release sourcemaps to Sentry for "${config.name}" build`)
       })
+
+      // Add WebpackPlugin to latest build config
+
+      const config = webpackConfigs[webpackConfigs.length - 1]
+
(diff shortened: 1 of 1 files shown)


lib/module.js
```diff
@@ -164,24 +164,30 @@ export default function SentryModule (moduleOptions) {
 
   // Enable publishing of sourcemaps
   if (!options.disabled) {
-    this.extendBuild((config, { isClient, isModern, isDev }) => {
-      if (!options.publishRelease || isDev) {
+    this.nuxt.hook('webpack:config', (webpackConfigs) => {
+      if (!options.publishRelease || this.options.dev) {
```
rchl: Using `this.options.dev` would change the logic. In old logic the WebpackPlugin was only added in one of the configs while with this it will be added to all configs. That would result in publishing release on building every configuration (server, client, modern).

rchl: Hmm, actually that's not true. I guess it should be equivalent to old code as the configs should have devmode set to the same value as nuxt config.

rchl: But then we don't even have to set up the hook if that first line of conditions is true.

matthieusieben: `isDev` is actually defined using `this.options.dev`

https://github.com/nuxt/nuxt.js/blob/e8aca9eb117851047e82e94948ff8b4bcb464b1a/packages/webpack/src/config/base.js#L38

matthieusieben: > But then we don't even have to set up the hook if that first line of conditions is true.

right

matthieusieben: fixed

lib/module.js
```diff
@@ -164,24 +164,30 @@ export default function SentryModule (moduleOptions) {
...
         return
       }
 
-      config.devtool = options.sourceMapStyle
+      webpackConfigs.forEach((config) => {
+        config.devtool = options.sourceMapStyle
```
matthieusieben: I would suggest to move this line after the exit conditions (at the end) as there is no need to compute the source map if they won't be used...

matthieusieben: What do you think @rchl ?

rchl: I think it's needed here since we need to generate source maps for client configs so that when building server we publish with source maps (also for the client configs).

lib/module.js
```diff
@@ -163,25 +163,28 @@ export default function SentryModule (moduleOptions) {
...
+  if (options.publishRelease && !options.disabled && !this.options.dev) {
+    this.nuxt.hook('webpack:config', (webpackConfigs) => {
+      webpackConfigs.forEach((config) => {
+        config.devtool = options.sourceMapStyle
+
+        const isClient = config.name === 'modern' || config.name === 'client'
+        const isModern = config.name === 'modern'
+
+        // when not in spa/universal mode upload only at server build
+        if (isClient && this.options.mode !== 'spa' && this.options.mode !== 'universal') {
```
rchl: Why was `this.options.mode !== 'universal'` condition added here?
I'd say we want to return here for client config in universal mode.

matthieusieben: Humm, I realized with the change in the log that the module was not uploading my client files to sentry, which is why I added this change. Why woudn't we want to send map files for client build to sentry ?

rchl: Nuxt builds all configs sequentially. First `client` then `modern` then `server`. We only want to add the sentry webpack plugin on the `server` stage since at that point client and server code is built. We don't want to upload when building clients as that would be overwritten later on server stage anyway.

Of course, when building SPA-only, we need to upload on client stage already as there will be no server to build.

rchl: Ping

matthieusieben: Are you sure that the WebpackPlugin will upload files that were not processed by the plugin ? this seems kinda odd to me. Anyway, if that is the case, I would change the code to directly add the plugin to the last config (without assuming any build order):

```js
webpackConfigs.forEach((config) => {
  config.devtool = options.sourceMapStyle
})

const config = webpackConfigs[webpackConfigs.length - 1]

config.plugins = config.plugins || []
config.plugins.push(new WebpackPlugin(options.webpackConfig))

logger.info(`Enabling uploading of release sourcemaps to Sentry for "${config.name}" build`)
```

rchl: > Are you sure that the WebpackPlugin will upload files that were not processed by the plugin ?

Plugin is told to upload stuff from `.nuxt/dist/server` and `.nuxt/dist/client`. On client building, it would not have server bundle and maps for it yet, so we postpone it until the last stage.

> Anyway, if that is the case, I would change the code to directly add the plugin to the last config (without assuming any build order):

That sounds good as long as we can rely on order of configs matching build order.

matthieusieben: I think it's a weird usage for a Webpack plugin linked to a build to manipulate code from another build. Especially since the `options.build.parallel` option might break any assumption we make on the order of execution anyway. See [here](https://github.com/nuxt/nuxt.js/blob/903c3de5b65edb1c3eed763371a72c1b534e22f7/packages/webpack/src/builder.js#L100-L102).

As I see it, either you use add `SentryWebpackPlugin` for every webpack config, or you use the `build:compile` hook to detect when the build is over, and use `@sentry/cli` directly to upload the release.

```js
const configNames = []
this.nuxt.hook('webpack:config', (webpackConfigs) => {
      webpackConfigs.forEach((config) => {
        config.devtool = options.sourceMapStyle
        configNames.push(config.name)
      })
})

this.nuxt.hook('build:compiled', ({ name, compiler, stats }) => {
  const idx = configNames.indexOf(name)
  if (idx !== -1 && configNames.splice(idx, 1)  && configNames.length === 0) {
    // Publish the release using @sentry/cli
  }
})
```

matthieusieben: Or you prevent parallel build:
```js
    if (options.build.parallel) {
      throw new Error('@nuxtjs/sentry is not able to publish a release in parallel build mode.')
    }
```

rchl: I'm certainly open to investigating a better solution. Either by you or myself when I have time.

Using `cli` might be a solution but probably requires extra dependency (how big?) and would be a bit awkward to run from the code maybe (unless it has programmatic interface also)?

matthieusieben: Actually i am wrong, parallel build is only for dev, for which publish is disabled...

matthieusieben: https://www.npmjs.com/package/@sentry/webpack-plugin?activeTab=dependencies is actually a dependency of webpack plugin

matthieusieben: Anyway, you can keep this in mind as an improvement to this module's code base. My latest commit to this PR is not really affected by this.

lib/module.js
```diff
@@ -163,24 +163,19 @@ export default function SentryModule (moduleOptions) {
   }
 
   // Enable publishing of sourcemaps
-  if (!options.disabled) {
-    this.extendBuild((config, { isClient, isModern, isDev }) => {
-      if (!options.publishRelease || isDev) {
-        return
-      }
+  if (options.publishRelease && !options.disabled && !this.options.dev) {
+    this.nuxt.hook('webpack:config', (webpackConfigs) => {
```
rchl: I've just check and this hook was introduced in [v2.10.0](https://github.com/nuxt/nuxt.js/pull/6468).
That's pretty recent. I'm not sure what to do about that. It might require major version change and note in breaking changes...

lib/module.js
```diff
@@ -163,24 +163,19 @@ export default function SentryModule (moduleOptions) {
...
+      webpackConfigs.forEach((config) => {
```
rchl: I generally prefer for/of loops. No unnecessary function calls then.
It's just a note though as the behavior is the same.

lib/module.js
```diff
@@ -163,24 +163,19 @@ export default function SentryModule (moduleOptions) {
...
+        config.devtool = options.sourceMapStyle
+      })
 
-      config.devtool = options.sourceMapStyle
+      // Add WebpackPlugin to latest build config
```
rchl:
```suggestion
      // Add WebpackPlugin to last build config
```

lib/module.js
rchl: Yeah, I'll have to go with major release.

rchl:
> If a custom webpack config is added by another module, that other config would **not** have the `WebpackPlugin` added

Just FYI: While this fix makes it possible, one also has to be aware that module registration order plays a role here and it will only work if other module runs first.
I don't like that about Nuxt -- should have some explicit dependency/priority system.

rchl:
Thank you for working on that. It's much nicer now. Hopefully works fine too :)

PR #169, opened 2020-03-24
chore(release): use release-it for making releases

rchl:
Main advantage being that it:
 - allows reviewing changelog before release
 - creates Github release (when GITHUB_TOKEN is provided)

Kept previous release package.json script renamed to `release-legacy`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "main": "lib/module.js",
   "scripts": {
     "lint": "eslint --ext .vue,.js lib test",
-    "release": "standard-version && git push --follow-tags && npm publish",
+    "release": "release-it",
     "test": "npm run lint && jest"
   },
   "babel": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,9 +16,29 @@
   "main": "lib/module.js",
   "scripts": {
     "lint": "eslint --ext .vue,.js lib test",
-    "release": "standard-version && git push --follow-tags && npm publish",
+    "release": "release-it",
+    "release-legacy": "standard-version && git push --follow-tags && npm publish",
     "test": "npm run lint && jest"
   },
+  "release-it": {
```
pi0: I would suggest moving it to `.release-it.js`:

- This configuration does not needs to be published with package. `package.json` is always in NPM package
- It is easier to update and maintain to be consistent across all module repos -- we can even create a preset in future

package.json
```diff
@@ -16,9 +16,29 @@
   "main": "lib/module.js",
   "scripts": {
     "lint": "eslint --ext .vue,.js lib test",
-    "release": "standard-version && git push --follow-tags && npm publish",
+    "release": "release-it",
+    "release-legacy": "standard-version && git push --follow-tags && npm publish",
```
pi0: Is it required to keep `standard-version` and legacy script? IMO if anything goes wrong we can simply revert back :)

rchl: I'm fine with removing. I thought I might annoy other maintainers if I force a new way on them.

PR #174, opened 2020-04-23
Leave possibility to fill manually `urlPrefix` field for windows users

VRuzhentsov:
Because currently upload filename looks like `\_nuxt\/001c573c077135db8d5a.js`
And I want to set `urlPrefix: "~/_nuxt/"` by myself

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -1,4 +1,4 @@
-import { resolve, join } from 'path'
+import { resolve, posix } from 'path'
 import consola from 'consola'
 import deepMerge from 'deepmerge'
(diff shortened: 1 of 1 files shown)


rchl:
That looks like a bug and we shouldn't need a separate option for it, just fix the bug.
Basically we should not use `path.join` as that is platform-specific.

rchl:
Or rather use `path.posix.join` which uses forward slashes always.

VRuzhentsov:
Pushed with `posix`

rchl:
Thank you.

PR #186, opened 2020-06-03
fix: log state of the Sentry logging properly

### LINKED ISSUE #183: Contradicting log messages
When running sentry-module in dev mode it first logs:

> Errors will not be logged because no DSN has been provided

but immediately after that it logs
> Started logging errors to Sentry

See also: 
image

So which statement is true? ;)

rchl:
Resolves #183

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ Normally, setting required DSN information would be enough.
 ### initialize
 - Type: `Boolean`
   - Default: `process.env.SENTRY_INITIALIZE || true`
+  - Can be used to add the `$sentry` object without initializing it, which will result in not reporting errors to Sentry when they happen but not crashing on calling the Sentry APIs.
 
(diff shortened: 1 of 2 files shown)


rchl:
I knew that. The thing is barely noticeable and easy to miss.
But I can remove it after all. I'm going back and forth on that.

pimlie:
Maybe we could change the first part of the message to: `Errors won't be reported to Sentry because ${why}`

rchl:
What do you think about what I did: `Sentry errors will not be logged because ...`?
I guess it's a bit broken grammar but my intention was to have "Sentry" at the beginning so that it's easy to scan where is the message from.

pimlie:
What about `Sentry reporting is disabled because ...` or someting similar? Otherwise people could maybe think its just about errors caused by Sentry itself.

PR #187, opened 2020-06-03
fix: add an option to toggle logs when calling mocked instance

### LINKED ISSUE #176: How to programatically check if Sentry is disabled?
Before, I used to rely on `this.$sentry` being undefined, but a recent update seems to have broken that and I'm getting a warning with `$sentry.captureException() called, but sentry plugin is disabled.` How can I check if Sentry is disabled and suppress the warning?

rchl:
Resolves #176

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 !.eslintrc.js
 !.release-it.js
 lib/sentry.client.js
+lib/sentry.mocked.js
 lib/sentry.server.js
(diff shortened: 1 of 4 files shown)


rchl:
@simllll

simllll:
Thanks for tagging! To be honest: we liked this additional logging very much. It only outputs something if sentry is called - which is only in an error scenario in our case. It helps a lot to see what went wrong in dev, or if something is intended to fail if it is sent to sentry or not (which is sometimes the case and sometimes not).

I would vote for an additional option to let sentry be verbose when disabled. But as a trade off, it can be disabled by default?

rchl:
I've added an option to log mock calls (enabled by default to not create a breaking change).

PR #196, opened 2020-07-01
Sentry Perfomance Monitoring

Manfies:
Hi. There is [Sentry Performance Monitoring](https://docs.sentry.io/performance-monitoring/getting-started/?platform=vue) beta feature, that required Tracing integration from `@sentry/apm`. I have added import from APM integrations, if there is `amp: true` option. It works perfectly for me, but can you review?

Screenshot 2020-07-01 at 14 14 31


rchl:
You need to remove the package-lock.json that you've added.
Use yarn for this module.
Also, we need `@sentry/apm` dependency I suppose but we probably shouldn't add it unconditionally for everyone.
Probably the best course of action is to just mention in the documentation that it needs to be installed when using apm integration.

lib/module.js
```diff
@@ -99,6 +99,16 @@ export default function SentryModule (moduleOptions) {
 
   // Register the client plugin
   if (!options.disabled && !options.disableClientSide) {
+    const apmIntegrations = {}
+    const integrationsKeys = Object.keys(options.clientIntegrations)
+    integrationsKeys.forEach((integrationName) => {
+      const integration = options.clientIntegrations[integrationName]
+      if (integration.apm) {
+        apmIntegrations[integrationName] = integration
+        delete options.clientIntegrations[integrationName]
```
rchl, line 108: We'd need this documented also but before we do, we should decide on the best way forward.

I think this module should make it super easy to enable tracing.

Right now one, per the example in documentation:

```js
import * as Sentry from "@sentry/browser";
import { Vue as VueIntegration } from "@sentry/integrations";
import { Integrations } from "@sentry/apm";
import Vue from "vue";

Sentry.init({
  // ...
  integrations: [
    new Integrations.Tracing(),
    new VueIntegration({
      Vue,
      tracing: true
    })
  ],
  tracesSampleRate: 1
});
```

has to do three things:
 - enable `tracing: true` for the Vue integration
 - add Tracing integration
 - set `tracesSampleRate` option.

All those can be done currently manually, I guess, but I think we should make it simpler. Possibly just a separate object option for this like:

```
enableTracing: {
   tracesSampleRate: 1,
   options: {}  // Optional options for the tracing integration
}
```

rchl:
Also, I wonder about the server-side... What makes sense here? Should we enable Vue tracing on the server also?

lib/module.js
Manfies: @rchl I think it makes sense to use `enableTracing` option, because there are only two Integration inside APM - Tracing and Expess (which is not needed for nuxt). I think i can extend this PR for this option.

rchl: I'm just having second thoughts about what will happen when they add another "performance monitoring" integration that we'll want to enable...

But it probably will still make sense to have tracing-specific configuration.

lib/sentry.client.js
```diff
@@ -1,5 +1,6 @@
 import VueLib from 'vue'
 import * as Sentry from '@sentry/browser'
+import { Integrations as ApmIntegrations } from "@sentry/apm"
```
rchl, line 3: I think we don't want to include this dependency unconditionally which means we have to use dynamic import and import it only when options say so.

Manfies:
@rchl - yes, i also saw for integrate tracing for server side, but decided to use only Vue tracing for now. Judging by the documentation, only the HTTP module is needed to trace the server side? Right? If true, then, as I understand it, can this be configured with the current version of the module?

lib/sentry.client.js
Manfies: You are right.

EdisonHarada:
Any idea on when this feature will be merged?

rchl:
When it's finished. There are still open issues.

gedhean:
How could I help to get this merged? 

Also, I would like to add APM support for the server.

rchl:
#198 is refactoring a lot of code and is almost done so we can revisit after that's merged.

lucianholt97:
@rchl are there any updates on this since #198 was merged?

dannywp:
Any updates on this?

rchl:
Feel free to contribute by implementing it. Can base your change on this PR, taking into account the feedback given here.

I'm not sure when I will have time to work on it myself.

rchl:
Superseded by #250

Issue #127, opened 2019-11-28
Lazy loading of @sentry/browser


ackushiw:
### What problem does this feature solve?

This will reduce the app bundle size and improve initial loading speeds.

### What does the proposed changes look like?

Following the guide here https://docs.sentry.io/platforms/javascript/#lazy-loading-sentry it should be possible, though it looks like the sentry loader is  not available via npm :/

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c77">#c77</a>)</em></sub></div>

Atinux:
Actually they they to use the `defer` attribute to lazy load, every scripts added by Nuxt already use the `defer` attribute :)

rchl:
While `defer` attribute is not exactly comparable to lazy loading (it means the script will be parsed after the page is parsed but still before DOMContentLoaded), using lazy-loaded loader would have major drawback that you wouldn't have the guarantee that Sentry catches all the errors. And of course it's not currently possible to integrate it (even as an option) as it's meant for browser-only usage and not through npm so webpack is not able to bundle it.

rchl:
The question is... what would such lazy-loaded Sentry provide comparing to the one that is "integrated" (bundled through their SDK via this module)?

I'm pretty sure it won't be as well integrated as SDK.
 - It won't be available in the Vue context using `this.$sentry.
 - It won't have Vue-specific integrations listed in https://docs.sentry.io/platforms/javascript/vue/.

Isn't that a deal-breaker? You would have to be quite careful when using `this.$sentry` not to use it on the client.

We could still pursue this path but, if added, there would have to be a big note saying how this affects everything.

pimlie:
It should be possible to use a Nuxt plugin and pass the references we need to sentry only once ready. From their sdk docs they dont seem to require that you bind to Vue's errorHandler before your app is mounted.
Within the Nuxt plugin you would wait until nuxt is ready and only then do a dynamic import & initilization.

For `this.$sentry` we could add a ready method that you have to await, a bit similar to how $nextTick works. This would be a breaking change, but it would result in your browser not having to download or parse or execute a chunk of 20KB of gzipped javascript.

```js
  await this.$sentry.ready()
  this.$sentry.whatever()
}, 1)
```

Possibilities on how to implement this are probably endless. I do agree that preferably this would be totally solved by sentry themselves though

[...]

PR #198, opened 2020-07-04
feat: support async/delayed loading of Sentry

pimlie:
Resolves: #127 

This pr adds a config option `async`. If `true` then sentry will only be loaded when `window.onNuxtReady` has been called.

> Im not totally sure yet whether using onNuxtReady is sufficient, it would be better if the user can determine by himself when he is ready to load sentry but not sure how such an API would look like. Suggestions are welcome!

Looking at the existing mocked client, the only usable API that sentry/browser exports are functions. So on build I retrieve that list and then on plugin init I create a mock for all calls to those functions. Then when sentry is really loaded all these calls are repeated on the actual Sentry object.

I dont think it has much use to also listen to the Vue global error handler, I didnt test it but I expect those errors to bubble up to `window.onerror` and I add a listener for that.

Also did some general cleanup.


rchl:
Awesome.
I'm on vacation so the review might take a bit longer but I will get to it eventually.

Just from a cursory look I would suggest:
- maybe call the option `lazy` as that seems more high level and use friendly term I think
- document the new option. Ideally mentioning also the downsides

pimlie:
:+1: for using `lazy` instead

- Have added a `$sentryLoad` callback so users could explicitly define when they want to load sentry.
- Also added some more options to control lazy loading behaviour

See the updated readme for all available options and how they work

README.md
```diff
@@ -80,6 +80,147 @@ async nuxtServerInit({ commit }, { $sentry }) {
 }
 ```
 
+## Lazy Loading (on the client)
```
rchl, line 83: Maybe this section should be after the `Options` section? Options are probably the first thing (after "setup") that people should look at and it refers to more details "lazy loading" section.

pimlie: To me options is more like a reference, while Lazy Loading is more part of the docs explaning functionality. Because of that I used this order.

Eg if we would have a vuepress website for the docs then Lazy Loading would be under Guide but Options would be under API. Agree that the lazy options could be moved down maybe but thought it would be better to keep all Lazy loading stuff together.

rchl: OK. Do what you prefer.

lib/module.js
```diff
@@ -95,38 +96,83 @@ export default function SentryModule (moduleOptions) {
...
+    const defaultLazyOptions = {
+      injectMock: true,
+      injectLoadHook: false,
+      mockApiMethods: true,
+      chunkName: 'sentry',
+      webpackPrefetch: false,
+      webpackPreload: false
+    }
+
+    options.lazy = Object.assign({}, defaultLazyOptions, options.lazy)
```
rchl, line 109: I'd prefer to have all options under the initial default object.
I guess you did it this way to shave some bytes when lazy loading is not used but I don't think it's worth it.
Also, it would make it problematic when annotating default options with types (which I have a local branch for).

README.md
pimlie: Have added a Configure section after Setup with a shortcut to Options, that way its still the first thing after setup in the readme :)

lib/plugin.lazy.js
```diff
@@ -0,0 +1,109 @@
...
+  ctx.$sentryLoad = loadSentryHook
+  <% } else { %>
+  window.onNuxtReady(loadSentryHook)
+  <% } %>
+
+  const sentryReady = () => {
+    return new Promise(resolve => {
+      resolveSentryReady = resolve
+    })
+  }
```
rchl, line 36: If `sentryReady()` is called more than once then each call will return a new Promise and only the last one will get resolved. So instead it should be something like:

```js
const sentryReadyPromise = new Promise(resolve => {
  resolveSentryReady = resolve
})

const sentryReady = () => sentryReadyPromise
```

lib/module.js
```diff
@@ -95,38 +96,83 @@ export default function SentryModule (moduleOptions) {
...
+  }
+
+  const apiMethods = await getBrowserApiMethods()
```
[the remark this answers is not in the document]
rchl: Or keep async but call it only when `lazy` is enabled by initializing `apiMethods` to an empty array outside of the `lazy` branch. And that should also join those two `if (options.lazy) {` blocks into one.

lib/module.js
```diff
@@ -199,3 +244,16 @@ export default function SentryModule (moduleOptions) {
     })
   }
 }
+
+async function getBrowserApiMethods () {
+  const SentryBrowser = await import('@sentry/browser')
+
+  const browserMethods = []
+  for (const key in SentryBrowser) {
```
rchl, line 252: Just a style thing but iteration using `const [key, value] of Object.entries(SentryBrowser)` would be nicer.

lib/plugin.server.js
```diff
@@ -0,0 +1,13 @@
...
+  // Inject Sentry to the context as $sentry
+  inject('sentry', sentry)
+  ctx.$sentry = sentry
+
+  <% if (options.lazy) { %>
+  const sentryReady = () => Promise.resolve(sentry)
+  inject('sentryReady', sentryReady)
+  ctx.$sentryReady = sentryReady
+  <% } %>
+}
```
rchl, line 13: Would it make sense to also mock `sentryLoad` function?
We could expect people to only call it from the client but it probably wouldn't hurt to also handle it on the server.

lib/module.js
```diff
@@ -95,38 +96,83 @@ export default function SentryModule (moduleOptions) {
...
+    }
+  })
+
   // Register the server plugin
-  if (!options.disabled && !options.disableServerSide) {
+  let pluginOptionServer
+  if (options.disabled || options.disableServerSide) {
+    pluginOptionServer = 'mocked'
+  } else {
+    pluginOptionServer = 'server'
```
rchl, line 174: With `disabled: true` (and `lazy` not set), we'll use "mocked" plugin for the server but we are not passing `apiMethods` and `logMockCalls` to it so things will crash.

(Can't mark the issue at proper line.)

(BTW. Enabling "mocked" plugin on the server is not really related to lazy loading but I guess it makes sense. Unless I'm missing something.)

lib/module.js
```diff
@@ -152,14 +193,18 @@ export default function SentryModule (moduleOptions) {
...
-      mode: 'server'
-    })
   }
 
+  this.addPlugin({
+    src: resolve(__dirname, `plugin.${pluginOptionServer}.js`),
+    fileName: 'sentry.server.js',
+    mode: 'server',
+    options: {
+      dev: this.options.dev,
```
rchl, line 203: I think it shouldn't be necessary to pass `dev` to server.

README.md
```diff
@@ -80,6 +83,147 @@ async nuxtServerInit({ commit }, { $sentry }) {
...
+      window.removeEventListener('error', errorListener)
+    }
+    window.addEventListener('error', errorListener)
+  },
+  destroyed() {
+    window.removeEventListener('error', this.errorListener)
+  }
+```
+
+### Lazy Options
```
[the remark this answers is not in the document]
pimlie: I dont mind, one of the main reasons why I kept this separate was because I didnt want to think about how to display second order options ;)

So, do you have a suggestion for that? Feel free to just commit directly into this pr otherwise

lib/module.js
pimlie: I have a love/hate relationship with Object.entries and in this case it feels way to verbose, it wont make the code better readable at all imo

lib/plugin.lazy.js
pimlie: Good catch, will fix!

lib/plugin.lazy.js
```diff
@@ -0,0 +1,109 @@
...
+  <% } %>
+
+  <% if (options.lazy.injectMock) { %>
+  window.removeEventListener('error', SentryMock.captureException)
+
+  delayedCalls.forEach(([methodName, args]) => Sentry[methodName].apply(Sentry, args))
+  <% } %>
+
+  // Inject Sentry to the context as $sentry
+  resolveSentryReady(Sentry)
```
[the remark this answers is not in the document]
pimlie: Actually the promise should always be set and not just until you call sentryReady. This should be fixed by the next issue

lib/plugin.server.js
pimlie: Im not a big fan of mocking functions that shouldnt be called anyway as it will only allow users to write faulty code and then hide that from them.

Can we think of a use-case where you'd actually want to use sentryLoad on a server? If not I think we could just omit it

lib/module.js
pimlie: Have added the missing options.

Enabling the mocked plugin is whats happening now already right? I'm quite sure I only refactored this logic

lib/module.js
pimlie: I think I added this to only log warnings in development mode but forgot to use it in the server plugin. Not sure if its a good idea or not, but I dont think you'd ever want to log `sentry.${key}() called, but Sentry plugin is disabled. Arguments` in production...

rchl: But then `logMockCalls` is used for deciding that so users can have any logic they want.

lib/module.js
rchl: Yeah, you're right. It was like that before already. And it seems like the `logMockCalls` handling was kinda broken as we didn't pass that option before either.

lib/plugin.server.js
rchl: I was thinking that someone could put the call in their plugin that runs on both the client and the server.

I don't have strong opinions on that so will just resolve. Let's say that user must explicitly do the client check.

lib/module.js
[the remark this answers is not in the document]
rchl: OK

README.md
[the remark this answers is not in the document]
rchl: I'll play with it and push a change if I can come up with a good solution.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,110 @@
+import VueLib from 'vue'
+
+<% if (options.lazy.injectMock) { %>
```
rchl, line 3: I absolutely hate templates in plugins as then both linting and type checking are not usable.
For nuxt-i18n I'm putting all options in a separate template and then importing those from the actual plugins. Of course you can't control if the code is inserted or not but typically control flows are enough.

I'm not saying that should be changed for this PR but I'll probably try that later if it's not.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,110 @@
...
+  // Dont unset delayedCalls & SentryMock during
+  // development, this will cause HMR issues
+  <% if (!options.dev) { %>
+  delayedCalls = undefined
+  SentryMock = undefined
+  <% } else { %>
+  delayedCalls = []
+  <% } %>
+  <% } %>
+  resolveSentryReady = undefined
```
rchl, line 109: This variable was not updated. I'll push a fix.

(This is why we need linting / type checking)

lib/module.js
rchl, line 112: Not feeling great about this one since `mockApiMethods` is part of the lazy options but `apiMethods` is also used when not using `lazy`. So you can only customize what is mocked when also using `lazy`.

lib/module.js
pimlie: Yeah, thats true. Just feels as a bad practice we should prevent to log such a warning message in prod. But am ok with leaving it as is

lib/plugin.lazy.js
pimlie: Or just better tests, we could just generate the plugin (and the other nuxt files) without building the files and then check that file. Have done something similar in my firebase module pr

lib/module.js
pimlie: :+1:

rchl: (I have not addressed this one, even though Github thinks I did)

What do you think about removing the `mockApiMethods` option and always mocking all methods if `injectMock` is enabled? It feels to me too eager to add `mockApiMethods` but maybe that's because I don't see the use case for only enabled mocking of some methods.

pimlie: Its a (micro-)optimization for your bundle size, ie why inject those method names into your bundle when you know you wont ever use them.

rchl: It's just a dozen of function names so I don't feel like it's worth adding an option for that.

And if kept, it would have to be refactored to be a global option and apply also to non-lazy mock.

pimlie: I implemented this for lazy only because only lazy really needs the mock/proxy methods. Also, Im a strong believer that any module should provide full configurability as to which options to include that affect bundle size. This single change might seem small, but its already 300bytes uncompressed that needs to be unzipped + parsed. If Im using 10 modules which would be this easy-going, thats 3KB. All in all its not about this single optimization but about all of these small optimizations combined that will provide a benefit.

In all other cases I actually would prefer to just not provide a mock. Eg if a user decides to not include sentry on the server then he is responsible himself to just not include it. Adding a mock for that is just too much user friendliness that only benefits bad practices. Or I might be misinterpreting why the mock existed originally.

lib/plugin.lazy.js
pimlie: Afaik templates can & will be dropped for Nuxt v3 at least ;)

lib/module.js
rchl: I think the mock was more for development vs. production cases. You would have mocked instance during development so that no issue would be reported.

But I can't recall what was the issue with setting `initialize: false` instead which should be able to satisfy that required too. Maybe @simllll can remind us his perspective on that.

README.md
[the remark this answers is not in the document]
rchl: Addressed with 0f8b060edfa8a2071d11d8a24cb7db97e18d8cca

lib/module.js
```diff
@@ -95,38 +96,81 @@ export default function SentryModule (moduleOptions) {
...
+      mockApiMethods: true,
+      chunkName: 'sentry',
+      webpackPrefetch: false,
+      webpackPreload: false
+    }
+
+    options.lazy = Object.assign({}, defaultLazyOptions, options.lazy)
+
+    if (!options.lazy.injectMock) {
+      options.lazy.mockApiMethods = []
```
rchl, line 114: A combination of `lazy: true` and server-side sentry being disabled would result in server-side not mocking the methods. Kinda unlikely case but still unexpected.

(I'll looking into making `mockApiMethods` a root option so this is just to not forget the issue.)

pimlie: @rchl Thanks man, really appreciate your help with this pr. Hope you don't mind it, but as you are opinionated about a lot of things (which is definitely a good thing!) it could be the fastest then trying to explain to me what you mean exactly :)

:bowing_man:

rchl: I've misunderstood how it works. I thought that "lazy" logic will affect what is mocked for the non-lazy mocks. But it's not the case so I'll just abandon my idea of making `mockApiMethods` a root option.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,109 @@
...
+  <% if (options.lazy.injectMock) { %>
+  window.removeEventListener('error', SentryMock.captureException)
+
+  delayedCalls.forEach(([methodName, args]) => Sentry[methodName].apply(Sentry, args))
+  <% } %>
+
+  // Inject Sentry to the context as $sentry
+  sentryReadyResolve(Sentry)
+  inject('sentry', Sentry)
+  inject('sentryReady', () => Promise.resolve(Sentry))
```
rchl, line 85: It's not possible to re-inject something that was already injected - see https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/vue-app/template/index.js#L185-L188

This will do nothing which is a problem because the potential code will keep using the mock (and crash) even after sentry has loaded.

pimlie: Ah damn, I should've know that. Its easily solvable except for the mock part. Either we should make sentry a function too or just skip the mock and tell people to always call+await `sentryReady`.

Eg we could reuse the mocked object reference, but then we have to assign the sentry moduel with `Object.assign(SentryMock, Sentry)`. That feels a bit hacky, but on the other side if it works it works. Wdyt?

pimlie: I found another way, we can just temporarily inject the SentryMock ourselves without calling Nuxt's inject. Then if the real Sentry is ready we use the inject call

pimlie:
To be sure I think we should release a beta version first before releasing this as stable

lib/plugin.lazy.js
```diff
@@ -107,3 +99,22 @@ async function loadSentry(ctx, inject) {
...
+// so we can later inject the final value
+// This method is adapted from the inject method in nuxt/vue-app/template/index.js
+function tempInject({ app }, key, value) {
+  key = '$' + key
+  // Add into app
+  app[key] = value
+  // Add into context
+  if (!app.context[key]) {
+    app.context[key] = value
+  }
```
rchl, line 113: It still looks to me like original `inject` won't be able to override that later:\
https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/vue-app/template/index.js#L177-L179

lib/plugin.lazy.js
```diff
@@ -79,21 +80,12 @@ async function loadSentry(ctx, inject) {
...
-  <% if (options.lazy.injectLoadHook) { %>
-  // Set noop for sentryLoad so we dont
-  // try to load Sentry more then once
-  const noop = _ => _
-  inject('sentryLoad', noop)
-  ctx.$sentryLoad = noop
-  <% } %>
 
+  sentryReadyResolve(Sentry)
+  sentryReadyPromise = Promise.resolve(Sentry)
```
rchl, line 86: I'm not sure I understand why it's necessary to reassign this variable with a new Promise.

lib/plugin.lazy.js
rchl: Those approaches sound hacky indeed.
Not sure what to do... A Proxy object would be the best but it's not supported by IE11.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,120 @@
...
+  <% if (options.initialize) { %>// Initialize sentry
+  const config = <%= serialize(options.config) %>
+  config.integrations = [
+    <%= Object.entries(options.integrations).map(([name, integration]) => {
+      if (name === 'Vue') {
+        return `new ${name}({Vue: VueLib, ...${serialize(integration)}})`
+      }
+
+      const integrationOptions = Object.entries(integration).map(([key, option]) => {
+        typeof option === 'function'
```
rchl, line 65: Needs a `return`.

lib/plugin.lazy.js
pimlie: Yeah, will delete that one before injecting again

lib/plugin.lazy.js
pimlie: Yeah, me neither. But I have seen cases where an await failed because there was no Thenable but just some var. So better save then sry.

lib/plugin.lazy.js
pimlie: Yeah, Proxies are a no go imo until Vue3. I think this solution will work without having to change `$sentry` to a function

lib/plugin.lazy.js
rchl: Still don't understand. `sentryReadyPromise` is a promise from the beginning. It's just reassigned to another promise here.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,121 @@
...
+// so we can later inject the final value
+// This method is adapted from the inject method in nuxt/vue-app/template/index.js
+function tempInject({ app }, key, value) {
+  key = '$' + key
+  // Add into app
+  app[key] = value
+  // Add into context
+  if (!app.context[key]) {
+    app.context[key] = value
+  }
```
rchl, line 114: What about this part of the original inject: https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/vue-app/template/index.js#L191-L199

This `tempInject` doesn't inject the getter to Vue prototype. Seems like a bit of a problem?

Can we instead do something like:

```js
function customInject({ app }, originalInject, key, value) {
  key = '$' + key
  delete app.context[key]
  originalInject(key, value)
}
```

lib/plugin.lazy.js
pimlie: Ah ok, now I understand what you are getting at. It probably doesnt matter much indeed as the promise is already resolved, but I like the explicitness of just overwriting it. If you want I can remove it

lib/plugin.lazy.js
pimlie: It does, it just does it without calling a plugin: https://github.com/nuxt-community/sentry-module/blob/1cd8da3648c775322d659a69d580c882dbd311aa/lib/plugin.lazy.js#L119 Afaik using the plugin is only useful as a preventive measure for only setting the property once.

You cant call Nuxt's inject twice because the property is added using defineProperty without configurable set so you after defining it once you wont be allowed to change it anymore.

lib/plugin.lazy.js
rchl: I'd say that it's enough to just call `sentryReadyResolve` to resolve original promise.

lib/plugin.lazy.js
rchl: The getter that original `inject` creates returns `this.$root.$options[key]` so it's fine if it's just added once, on first inject. The getter will return the most recent value assigned to `app[key]` (I assume that this is what `this.$root.$options[key]` refers to. So I think that my suggestion still makes sense to use both for the temp and later inject (or just the later one).

rchl: My point is to have smaller and more concise code and also reusing as much as possible of original inject. It seems that we really only need to force-inject key to `app.context`. Otherwise the original `inject` should all the rest of the work.

pimlie: @rchl Could you please try to change this yourself maybe and make sure that it works? You are probably perfectly right but I dont have time to keep looking into finding the ultimate perfect solution unfortunately ;)

rchl: Not immediately but yes, I can.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,121 @@
...
+  <% if (store) { %>
```
rchl, line 115: Remainder for myself: `store` is not defined.

rchl:
A crash in `mounted` hook won't be caught because it doesn't trigger global error handler.

For example:
```js
    mounted() {
        crashme();
    },
```

I wonder if anything can be done about that by hooking more deeply into Vue error handling.
Even if there is a note saying "some errors won't be caught", it's still quite unfortunate to miss those potentially very common cases.

rchl:
> Who is responsible for cathcing the error, is that Nuxt or Vue?

Vue.

The lifecycle hooks are called with this code in `vue.runtime.*.js`:
```js
function invokeWithErrorHandling (
  handler,
  context,
  args,
  vm,
  info
) {
  var res;
  try {
    res = args ? handler.apply(context, args) : handler.call(context);
    if (res && !res._isVue && isPromise(res) && !res._handled) {
      res.catch(function (e) { return handleError(e, vm, info + " (Promise/async)"); });
      // issue #9511
      // avoid catch triggering multiple times when nested calls
      res._handled = true;
    }
  } catch (e) {
    handleError(e, vm, info);
  }
  return res
}
```

But that then calls:
```js
function handleError (err, vm, info) {
  // Deactivate deps tracking while processing error handler to avoid possible infinite rendering.
  // See: https://github.com/vuejs/vuex/issues/1505
  pushTarget();
  try {
    if (vm) {
      var cur = vm;
      while ((cur = cur.$parent)) {
        var hooks = cur.$options.errorCaptured;
        if (hooks) {
          for (var i = 0; i < hooks.length; i++) {
            try {
              var capture = hooks[i].call(cur, err, vm, info) === false;
              if (capture) { return }
            } catch (e) {
              globalHandleError(e, cur, 'errorCaptured hook');
            }
          }
        }
      }
    }
    globalHandleError(err, vm, info);
  } finally {
    popTarget();
  }
}
```
which calls `globalHandleError`:
```js
function globalHandleError (err, vm, info) {
  if (config.errorHandler) {
    try {
      return config.errorHandler.call(null, err, vm, info)
    } catch (e) {
      // if the user intentionally throws the original error in the handler,
      // do not log it twice
      if (e !== err) {
        logError(e, null, 'config.errorHandler');
      }
    }
  }
  logError(err, vm, info);
}
```

So it looks like it should be possible to hook into that.

lib/plugin.lazy.js
rchl: It's actually quite hard to override keys injected to the Vue instance since what happens now is that those are injected to `app` object and then Vue instance is created with those options passed in and then `app` keys are copied into the instance's `$options` object. So changing Vue.prototype later will not affect those options.

pimlie: See my latest commits, this should work (it does with `yarn nuxt test/fixture/with-lazy-config/`). In theory it would be possible that Sentry is loaded at a later time then requested, because if you call sentryLoad before App.create then you have to wait until window.onNuxtReady. I added a warning about that in dev mode, but I tried and I couldnt even trigger that warning. Also it doesnt make any sense at all, because if you need that then you just shouldnt use lazy mode anyway. Therefore I think this edge-case is covered enough.

lib/plugin.lazy.js
pimlie: Shoudl be fixed now

pimlie:
Also added a Vue errorHandler, but thats only usefull when `lazy.injectMock = true`

lib/plugin.lazy.js
```diff
@@ -8,6 +8,18 @@ let SentryMock = {}
...
+const vueErrorHandler = VueLib.config.errorHandler
+
+VueLib.config.errorHandler = (err, vm, info) => {
+  vm.$sentry.captureException(err)
+
+  if (vueErrorHandler) {
+    return vueErrorHandler(err, vm, info)
+  }
+}
+<% } %>
```
rchl, line 21: Bbased on my previous reading of Sentry code, I think that will be problematic because when sentry is loaded, it will wrap our handler with its own, and then `captureException` will be triggered twice, once from here and once from Sentry.

We would need to restore state of `errorHandler` to what we started with before Sentry is loaded.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,139 @@
...
+  ctx.app[key] = value
+  // Add into context (always overwrite)
+  ctx[key] = value
+
+  <% if (store) { %>
+  // Add into store
+  ctx.store[key] = app[key]
+  <% } %>
+
+  window.<%= globals.nuxt %>.$options[key] = value
```
rchl, line 138: Nice!

lib/plugin.lazy.js
```diff
@@ -0,0 +1,139 @@
...
+function injectOverwrite(ctx, key, value) {
+  key = '$' + key
+  // Add into app
+  ctx.app[key] = value
+  // Add into context (always overwrite)
+  ctx[key] = value
+
+  <% if (store) { %>
+  // Add into store
+  ctx.store[key] = app[key]
```
rchl, line 135: Undefined variable `app` :).
Should be `value`.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,139 @@
...
+  inject('sentry', SentryMock)
+  <% } %>
+
+  const loadSentryHook = () => !loaded && loadSentry(ctx, inject)
+
+  <% if (options.lazy.injectLoadHook) { %>
+  inject('sentryLoad', loadSentryHook)
+  ctx.$sentryLoad = loadSentryHook
+  <% } else { %>
+  window.onNuxtReady(loadSentryHook)
```
rchl, line 41: `<%= globals.readyCallback %>`

lib/plugin.lazy.js
```diff
@@ -0,0 +1,139 @@
...
+  <% } else { %>
+  delayedCalls = []
+  <% } %>
+  <% } %>
+  sentryReadyResolve = undefined
+}
+
+
+// Temporary inject into Nuxt without using Nuxt's inject method,
+// so we can later inject the final value
```
rchl, line 124: This comment is outdated as now we are using this function to inject the final value.

But also, consider passing original `inject` to it, calling it and then just overwriting context[key] and $options as only those two really need custom handling.

lib/plugin.lazy.js
rchl: Thinking more about it, due to the nature of the global handler, I think it would be risky to try to unset our handler as it might have been changed by something else before Sentry does it.

A better idea would be to keep the function assigned but just change the inner-workings of it like:
```js
let captureExceptionHandler = (err, vm, info) => {
  vm.$sentry.captureException(err)
}

VueLib.config.errorHandler = (err, vm, info) => {
  captureExceptionHadler(err, vm, info)

  if (vueErrorHandler) {
    return vueErrorHandler(err, vm, info)
  }
}
```

and then, from `loadSentry()` re-assign `captureExceptionHandler` to a no-op function.

lib/plugin.lazy.js
```diff
@@ -0,0 +1,139 @@
...
+let delayedCalls = []
+let SentryMock = {}
+<% } %>
+let sentryReadyResolve
+let loaded = false
+
+<% if (options.lazy.injectMock) { %>
+const vueErrorHandler = VueLib.config.errorHandler
+
+VueLib.config.errorHandler = (err, vm, info) => {
```
rchl, line 14: One other thing I've just realized is that having the global handler set makes hides the errors from the console (see https://github.com/vuejs/vue/blob/7186940143704acc4ec046132f6a56e9c983e510/src/core/util/error.js#L59).

See how Sentry re-triggers the console message in its handler:
https://github.com/getsentry/sentry-javascript/blob/820bd77cca08cc6f4e209b6cbf8dcf82ab3f3477/packages/integrations/src/vue.ts#L405-L410

rchl: Addressed with a92de0a67cb5cecf125379c95bb2c77b48fdb2e2

lib/plugin.lazy.js
rchl: Addressed with 1eafbf2c0537640c98204654f8f1ebc23b0dde97

lib/plugin.lazy.js
rchl: Addressed with 1eafbf2

lib/plugin.lazy.js
rchl: Addressed with 95cd0ba7efa55d3ccabfd0f962cb9b71548ee06c

lib/plugin.lazy.js
rchl: Addressed with e5dc66434467207665ca115f258028965d233f50

rchl:
I've addressed all the issues I had with the code and it looks good to me now. You can re-check everything now. :)

lib/plugin.lazy.js
pimlie: @rchl You could wrap the sentry capture call in an if not loaded. Then at least sentry wouldnt be called twice?

pimlie:
Because you mentioned you didnt like that you couldnt lint the plugins: well you can. sort-of. We do this too for nuxt, just lint all the plugins from the test fixtures.

We just have to be aware that:
- somethings we cant rly lint correctly, ie anything that gets serialized. So just disable those whole sections
- if you dont test a plugin feature, it wont be included in any of the plugins so it could still contain a (lint) error

lib/plugin.lazy.js
rchl: It seems like using `loaded` would potentially leave some window where we might not catch errors since `loaded` is set at the beginning of `loadSentry` but loading might be delayed. So I think I'll introduce now one more variable for this.

rchl:
> We do this too for nuxt, just lint all the plugins from the test fixtures.

Sorry but this is not a good enough solution for me. The issues:
 - There is no immediate feedback when developing which slows down the process
 - For the linting to be complete the file would have to be generated and linted with all possible options combinations. Not realistic to do.

lib/plugin.lazy.js
rchl: Addressed in 4452ac3d86c9e07f11219f59e65d7ec5c7599931

pimlie:
> Sorry but this is not a good enough solution for me.

Sorry, but that doesnt make really sense to me. First of all because this linting works without any additional development required. That you cant lint it during development is a small downside, but you can correct it easily by performing 1 extra step.

If you want to put a lot of extra development time in this module, please improve test coverage. Before this PR there wasnt any realistic tests and frankly there still isnt. Proper testing is _WAY_ more important then any refactor that only benefits you and me as contributors to this module.

rchl:
> If you want to put a lot of extra development time in this module, please improve test coverage.

One doesn't have to exclude the other.

> That you cant lint it during development is a small downside, but you can correct it easily by performing 1 extra step.

It's that small of a downside if I can't see the issue while writing the code and will only know about it later when I run the tests. 
The tests might not even make it obvious what the error is so you need to spend even more time investigating the issue. It is a major productivity killer. 

While I agree that the tests are most important, it's way easier to develop with linter and language server help. This PR itself introduced issues a couple of times that would be caught early by linters and would never reach this PR and take both of our times.

Also, it's not fun to work with code like this:
Screenshot 2020-07-24 at 11 38 22

Even if I would disable linters locally to silence the invalid markers it's still a pain for the brain to process the code with templates. Even the simple things like where to insert braces and indentation are not obvious with templates.

So I think getting rid of templates is very important (it is for me at least).

pimlie:
@rchl Mostly valid points, but still I would recommend to not put time in this now but wait for nuxt v3 instead ;)

rchl:
Let me know if I have your final approval and I'll merge it.

pimlie:
@rchl Yes thanks, if you are ok then I am ok ;)

rchl:
You said to maybe release as a beta first but I think it's not necessary as non-lazy changes shouldn't have any issues and the lazy feature is new so it's OK if it has issues.

I'm not sure how much testing we would get when releasing as a beta. I fear that not much.

pimlie:
I will leave that to you to decide, but should be ok indeed.

Thanks for all your help!

rchl:
Thank you for the idea and implementation :)

pimlie:
I am wondering whether you feel a bit more confident that the most important errors are still catched & reported? Would you consider using lazy mode yourself?

rchl:
I do feel pretty confident about it myself and will switch to it in my company and see how things go.

I don't really think that client-side errors are that useful as there is so much crap caused by things out of our control that it's hard to keep track and understand all of them. So I'm not too worried about losing some reports. The ones within Vue itself should be working at least.

pimlie:
That was also exactly my reasoning, thanks for confirming ;)

patrickxchong:
Thanks for putting together this feature! I see that the README is updated with the lazy option but that's not reflected on the npm README. When will this be released on npm (either as an official release or as a beta)?

rchl:
@patrickxchong https://github.com/nuxt-community/sentry-module/issues/127#issuecomment-663826005

PR #203, opened 2020-07-23
feat: add private runtime config support

### LINKED ISSUE #202: Plugin not using runtime variables from $config
### What problem does this feature solve?

When using application in more than one environment, different configuration settings for sentry module are needed during runtime. 

### What does the proposed changes look like?

Override default configuration to use runtime config which would have higher priority than using environment variables.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c105">#c105</a>)</em></sub></div>

marosmola:
This PR resolves #202

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -53,7 +53,8 @@ export default function SentryModule (moduleOptions) {
     }
   }
 
-  const runtimeSentryConfig = this.nuxt.options.privateRuntimeConfig.sentry || {}
+  const privateRuntimeConfig = this.options.privateRuntimeConfig || {}
+  const runtimeSentryConfig = (privateRuntimeConfig && privateRuntimeConfig.sentry) || {}
   const topLevelOptions = this.options.sentry || {}
   const options = deepMerge.all([defaults, topLevelOptions, moduleOptions, runtimeSentryConfig])
 
--- a/lib/module.js
+++ b/lib/module.js
@@ -53,8 +53,9 @@ export default function SentryModule (moduleOptions) {
     }
   }
 
+  const runtimeSentryConfig = this.nuxt.options.privateRuntimeConfig.sentry || {}
   const topLevelOptions = this.options.sentry || {}
-  const options = deepMerge.all([defaults, topLevelOptions, moduleOptions])
+  const options = deepMerge.all([defaults, topLevelOptions, moduleOptions, runtimeSentryConfig])
 
   options.serverConfig = deepMerge.all([options.config, options.serverConfig])
   options.clientConfig = deepMerge.all([options.config, options.clientConfig])


rchl:
Apart from obvious problems like tests crashing and not taking into account Nuxt versions that don't support runtime config, this is also not correct because it will only apply to the server-side sentry. The correct way to do this that would also support client-side would be much more invasive.

If you are thinking of doing it properly then it should be done after #198 is merged as that one is rewriting a lot of code.

marosmola:
@rchl thank you for your reply. This was just example how this would work in my case because I am really missing this feature of runtime configuration an i think i would be beneficial for others also. 

>  If you are thinking of doing it properly then it should be done after #198 is merged as that one is rewriting a lot of code.

no problem

rchl:
#198 is integrated now.
I'm still doing some refactorings in #205 that will be quite major.

Just note that implementing this correctly is non-trivial. Currently, we are using templates to create the client-side plugins and those have all options hardcoded at build time. So changing that would require some work. I would probably go for separate options like `useRuntimeOptions` to toggle the behavior.

One thing that I don't see solution for is how to handle `publishRelease`. This happens at build-time and if you want the environment to be dynamic, for example, you would, in theory, have to upload releases for all possible environments when building.

lib/module.js
```diff
@@ -53,8 +53,10 @@ export default function SentryModule (moduleOptions) {
     }
   }
 
+  const privateRuntimeConfig = this.options.privateRuntimeConfig || {}
+  const runtimeSentryConfig = (privateRuntimeConfig && privateRuntimeConfig.sentry) || {}
```
IlyaSemenov, line 57: This is not going to work. `privateRuntimeConfig` is what is going to be available in generated plugin (in **run**-time), not in module (in **build**-time). You were probably misled by the fact that on developer machine, the environment is the same/similar during build and run. This is totally different in production.

You need to some like this in module:

```ts
  const defaults = {
    useRuntimeConfig: this.options.publicRuntimeConfig ? 'sentry' : undefined,
    ...
```

and this in plugin:

```ts
export default ({ app: { router }, $config }) => {

  const options = <%= JSON.stringify(options) %>
  <% if (options.useRuntimeConfig) { %>
    Object.assign(options, $config[options.useRuntimeConfig])
  <% } %>
  // You may want to remove `useRuntimeConfig` from the options object
  // if you're passing it directly to Sentry.
```

This will be compatible with Nuxt <2.13, and it will allow to override runtime config option key or disable it completely.

rchl:
Closing as stale. Follow-up in #254

PR #213, opened 2020-07-27
feat: add types for module configuration

rchl:


rchl:
@kevinmarrec maybe you want to have a look? Not necessarily from the module-specific point of view but more a generic TS types point of view.

types/sentry.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    disableClientSide?: boolean
+    disabled?: boolean
+    disableServerRelease?: boolean
+    disableServerSide?: boolean
+    dsn?: string | ''
+    initialize?: boolean
+    lazy?: boolean | LazyConfiguration
+    logMockCalls?: boolean
+    publishRelease?: boolean
+    repo?: string | ''
```
kevinmarrec, line 31:
```suggestion
    repo?: string
```

`''` is of type `string` 

image

types/sentry.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    serverConfig?: SentryOptions
+    serverIntegrations?: IntegrationsConfiguration
+    sourceMapStyle?: WebpackOptions.Devtool
+    webpackConfig?: SentryCliPluginOptions
+}
+
+type LazyConfigurationResolved = Required<LazyConfiguration>
+type ModuleConfigurationResolved = Required<ModuleConfiguration>
```
kevinmarrec, line 39: These types are not used ? Maybe you wanted to export them ?

rchl: Those are internal and I'm using them through JSDoc like so: https://github.com/nuxt-community/sentry-module/blob/bc65423448b391fa54c3322b19331f429e6deeea/lib/core/hooks.js#L30

I suppose I can create them dynamically in JSDoc comments instead.

rchl: I'm creating those in JSDoc now although still not sure which of the three possible options I should go with 👍 
 - declare internal here and not export
 - declare and export
 - create directly in jsdoc

kevinmarrec: Didn't know it was possible to use unexported types with JSDoc ^^

If only used in internals :
 - create directly in jsdoc
 - declare and export (but not export in main types entry a.k.a. index.d.ts)

Current state looks good to me

rchl: Thanks

PR #214, opened 2020-07-28
fix: enable console logging of Vue errors in development

rchl:
Sentry by default swallows all errors caught by Vue error handler. This
options re-posts them so that they show up in the console.

I have always had this enabled in my configuration and I think it makes sense to have it enabled by default.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -28,7 +28,7 @@ export default function SentryModule (moduleOptions) {
       ExtraErrorData: {},
       ReportingObserver: {},
       RewriteFrames: {},
-      Vue: { attachProps: true, logErrors: true }
+      Vue: { attachProps: true, logErrors: this.options.dev }
     },
     serverIntegrations: {
       Dedupe: {},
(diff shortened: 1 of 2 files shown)


lib/module.js
```diff
@@ -28,7 +28,7 @@ export default function SentryModule (moduleOptions) {
       ExtraErrorData: {},
       ReportingObserver: {},
       RewriteFrames: {},
-      Vue: { attachProps: true }
+      Vue: { attachProps: true, logErrors: true }
```
pimlie, line 31: Could we make it configurable anyway? That should be a quick fix and logging in production could be frowned upon..

rchl: Technically it is configurable by overriding `clientIntegrations`.
I wouldn't be opposed to adding a configuration option specifically for that if I would have a good place for it. I don't think that adding a new global option for configuration one specific integration would be worth it...

rchl: I don't really see any problems with posting errors in production as those are client errors and there is nothing to hide there.

But I'm open to setting the default value to `this.options.dev` instead.

rchl:
I've just realized that, since we are deep-merging options, it's easy to override settings for a specific integration but at the same time it's not possible to disable integration that is enabled by default.

But since no one has complained yet about that, we can leave as is.

pimlie:
As long as logging in prod is configurable then Im happy :)

PR #216, opened 2020-07-29
fix: crash on running module outside of git repo when release not set

### LINKED ISSUE #200: Getting sourcemaps / releases working
### Version

[v4.1.3](https://github.com/sentry-module/releases/tag/v4.1.3)

### Reproduction link

[https://github.com/cdnjs/static-website](https://github.com/cdnjs/static-website)

### Steps to reproduce

I have the following config in `nuxt.config.js`: 

```
    sentry: {
        dsn: sentryDsn,
        publishRelease: 'SENTRY_AUTH_TOKEN' in process.env,
        sourceMapStyle: 'hidden-source-map',
    },
```

`sentryDsn` is a valid DSN url imported from another config file.

The `SENTRY_AUTH_TOKEN` environment variable is set during build & start.

`.sentryclirc` has our org/project set correctly:

```
[defaults]
org = cloudflare
project = cdnjs-website
```

When `nuxt build` is run, it creates the release correctly on Sentry as far as I can tell, using the latest commit hash.

However, when errors are then generated in the app (after `nuxt start`), they appear in Sentry without any attached release and don't appear to be using the source maps.

### What is expected ?

Maybe I'm configuring something wrong, but I would expect errors generated both server-side and client-side to be correctly attached to the release generated during `nuxt build` and to use the source maps so that the stack trace is useful in Sentry.

Issues have no attached release and don't use the source maps, resulting in the stack trace being the raw bundled files which is unusable.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c104">#c104</a>)</em></sub></div>

rchl:
Catch Sentry CLI better to avoid crashing when running outside of git repo.

Resolves #200

### DIFF
--- a/lib/core/hooks.js
+++ b/lib/core/hooks.js
@@ -34,13 +34,14 @@ async function getBrowserApiMethods () {
 export async function buildHook (moduleContainer, options, logger) {
   if (!options.config.release) {
     // Determine "config.release" automatically from local repo if not provided.
-    // @ts-ignore
-    const SentryCli = (await import('@sentry/cli')).default
-    const cli = new SentryCli(null, { silent: true })
-    options.config.release = await cli.releases.proposeVersion()
+    try {
       // @ts-ignore
-      .then(version => `${version}`.trim())
-      .catch()
+      const SentryCli = (await import('@sentry/cli')).default
+      const cli = new SentryCli()
+      options.config.release = (await cli.releases.proposeVersion()).trim()
+    } catch {
+      // Ignore
+    }
   }
 
   options.serverConfig = deepMerge.all([options.config, options.serverConfig])


rchl:
I'm gonna go ahead with this as it's simple enough but feel free to raise any issue post-merge. :)

pimlie:
Yeah lgtm.

Was thinking that it could maybe be helpful to support `config.release` being a function? That way it would be a bit easier for ppl to integrate their own version scheme directly within their sentry options. So `config.release` could then be a bool, string or function?

rchl:
Maybe let's wait with that until someone asks?

pimlie:
Thats for you to decide, it shouldnt be that hard to add tho. Also as a module maintainer I think it does make sense in certain situation to add feature before ppl start asking about it. Determining a release version could be one of those (but happy to omit it for now, Im not using release versions myself anyway) :)

PR #219, opened 2020-07-29
fix: crash on iterating "delayedCalls"

rchl:
It looks like the code allowed "loadSentry()" to run more than once and
the first run would unset "delayedCalls" causing second run to crash on
trying to iterate it.

### DIFF
--- a/lib/plugin.lazy.js
+++ b/lib/plugin.lazy.js
@@ -7,9 +7,9 @@ let SentryMock = {}
 <% } %>
 let sentryReadyResolve
 let loadInitiated = false
+let loadCompleted = false
 
 <% if (options.lazy.injectMock) { %>
-let loadCompleted = false
 const vueErrorHandler = VueLib.config.errorHandler
 
 VueLib.config.errorHandler = (error, vm, info) => {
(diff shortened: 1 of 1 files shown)


rchl:
Hopefully handles this issue: https://sentry.io/share/issue/2c22b5e9282a4a7986d910433e84518f/#exception

There is quite a lot of extra checking in this PR that should stop the code running twice.

But I can't quite see how this linked error was possible as for `loadSentry` to trigger twice (with default lazy settings), we would need to add another `onNuxtReady` and while there was code that did that, that function could have only been triggered by `onNuxtReady` hook itself so `window.$nuxt` should have existed already...

rchl:
Hmm, I'm not even sure if that error is about calling `forEach` on `undefined` or about calling Sentry API from `forEach`. The error message is pretty pure on safari webview.

rchl:
Will get this in and see if it still reproduces.

rchl:
That didn't fix it...

I will have to inject some debugging code to try to figure it out.
I can only check it on production though so that makes it quite tricky to debug...

rchl:
Just realized that this is expected. It's reposting the error caught through error handler.
The part that made me confused is the stack trace pointing at the sentry plugin. I could probably repost from a timeout to hopefully lose the stack as it's not relevant in those cases.

PR #220, opened 2020-07-30
fix(lazy): improve error reporting for postponed global errors

rchl:
This ensures that the global errors that were delayed will be reported using the same mechanism (`onerror`) as they normally would. Makes it clearer in the Sentry UI what the error is.

### DIFF
--- a/lib/plugin.lazy.js
+++ b/lib/plugin.lazy.js
@@ -11,7 +11,13 @@ let loadCompleted = false
 
 <% if (options.lazy.injectMock) { %>
 let delayedGlobalErrors = []
-const delayGlobalError = error => delayedGlobalErrors.push(error)
+let delayedUnhandledRejections = []
+const delayGlobalError = function() {
+  delayedGlobalErrors.push([].slice.call(arguments))
+}
+const delayUnhandledRejection = function(event) {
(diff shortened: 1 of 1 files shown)


rchl:
After talking to Kamil Ogorek from Sentry, he pointed me at their own lazy-loading code at https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/loader.js#L135-L147 so I've decided that it makes sense to refactor our code to do thing in similar way. That is to call the `window.onerror` handler with saved arguments. Also added support for re-posting unhandled promise rejections like in their code.

rchl:
See their code at https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/loader.js#L135-L147

pimlie:
Sorry, I missed this PR.

+1 for capturing unhandled exceptions.

Calling window.onerror is possible because Sentry is listening to that I suppose? If so, then thats ok too ofc. But could you please refactor the code and abstract the delaying of the errors? Currently it has two variables & two for loops which are almost exactly the same, would be better to combine that into one so we have less code and a somewhat smaller bundle. Eg see the Sentry code you linked what already seem to do that :)

rchl:
Sentry does assign `window.onerror`, yes.

I'll put refactoring on my list of lower-priority things. :)
Feel free to handle that first if you have time.

Issue #230, opened 2020-09-13
serverSide not working when using Nuxt programatically


abakermi:
### Version

[v4.3.4](https://github.com/sentry-module/releases/tag/v4.3.4)

### Reproduction link

[https://github.com/abakermi/sentry-demo-production-mode](https://github.com/abakermi/sentry-demo-production-mode)

### Steps to reproduce

npm run build && NODE_ENV=production npm run start

### What is expected ?

$sentry is defined in fetch(),asyncData(),...

### What is actually happening?

$sentry={}

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/sentry-module/issues/c111">#c111</a>)</em></sub></div>

PR #231, opened 2020-09-13
support nodejs server with production environment

abakermi:
#230

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -49,7 +49,7 @@ export default function SentryModule (moduleOptions) {
       ],
       configFile: '.sentryclirc'
     },
-    isNodeServer :this.options.sentry.isNodeServer || false
+    isNodeServer: (this.options.sentry || {}).isNodeServer || false
   }
 
   const topLevelOptions = this.options.sentry || {}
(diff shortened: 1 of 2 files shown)


rchl:
- Undocumented new option
 - Not respecting `moduleOptions` (options can come from it also)
 - Doesn't seem to be specific to the production environment like fix description says but to programmatic Nuxt usage
 - I'd prefer this to be handled automatically. Would it work to always init on `ready`?

rchl (changes requested):
- Undocumented new option
 - Not respecting `moduleOptions` (options can come from it also)
 - Doesn't seem to be specific to the production environment like fix description says but to programmatic Nuxt usage
 - I'd prefer this to be handled automatically. Would it work to always init on `ready`?

abakermi:
@rchl  `ready` should work well , I will document the option

rchl:
There is no need to and the option in the first place if you are gonna use `ready` always.

rchl:
Using `ready` would mean that Sentry would also be initialized when building the project. Not sure about that myself but I'd lean towards it being OK. @pimlie @pi0 any opinions?

pi0:
Always using `ready` instead of `listen` (1) seems good idea  :+1:   We should just ensure to not double init on same node instance as `nuxt generate` with `target: 'static'` and cache (MISS) initializes nuxt twice once for build and once for generate.

Agree @rchl for going towards initializing sentry when building. But if not desired for any reason or is breaking change, we can use either `options._build` (2)  or `!options.server` to avoid initializing when building.

(1) I think `listenHook` name seems little bit misleading for source code as it [directly calls](https://github.com/nuxt-community/sentry-module/blob/c884b917249e784aaaae1a9d7bb186ba32df5d3a/lib/core/hooks.js#L211) `initSentry`

(2) As already mentioned, `options._*` flags are only prefixed to avoid conflicting with user options. You can safety rely on them :)

rchl:
Thanks @pi0 for your input. I'm addressing this and your comments in https://github.com/nuxt-community/sentry-module/pull/235.

PR #247, opened 2020-10-31
docs: update link to config options documentation

### LINKED ISSUE #246: Broken link in documentation
### Reproduction Link

here:

https://github.com/nuxt-community/sentry-module/blob/fe1c8d53f9c0b4c9f526cc3e260cbfc5a8134a60/README.md#L44

the link goes to https://docs.sentry.io/error-reporting/configuration/?platform=browser

which redirects to https://docs.sentry.io/platforms/browser showing 404 Not Found.

### What is Expected?

Non-broken link.

Broken link.

### Why is that important

The documentation never mentions if Sentry DSN in nuxt config should be public or private. I tried to follow the link to see the options, but it didn't work.

* On the one hand, public makes more sense because part of sentry integration runs on the client.
* On the other hand, using private DSN during SSR makes sense, too, and public could be derived by simply stripping the password part. Not sure if this module does that or not.

IlyaSemenov:
Fixes #246.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ See [Options](#options) for a list of available options
 
 ## Usage
 
-Enter your DSN in the Nuxt.js config file. Additional config settings can be found [here](https://docs.sentry.io/error-reporting/configuration/?platform=browser).
+Enter your DSN in the Nuxt.js config file. Additional config options can be found [here](https://docs.sentry.io/platforms/javascript/guides/vue/configuration/options/).
 
 ### Usage in Vue components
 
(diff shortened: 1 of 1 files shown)


rchl:
There is also a broken link in the description of the `config` option.

IlyaSemenov:
@rchl Right. I've updated the other link.

rchl:
BTW. Technically the https://docs.sentry.io/platforms/javascript/configuration/options/ wouldn't probably be more relevant here as it's not `Vue`-specific. But those pages are not really any different (at least now).

IlyaSemenov:
But doesn't Vue-specific page make more sense? After all, @nuxtjs/sentry is a plugin to Nuxt, and Nuxt is essentially Vue wrapper.

rchl:
The server-side part is technically Vue-free. Only the client-side configures the Vue integrations.

rchl:
Thanks

PR #250, opened 2020-11-10
feat: add support for performance monitoring/tracing

jonthemiller:
Based off the work of @Manfies in #196, but updated to take some of the feedback from the PR into account. 

* Added a top level `enableTracing` option, default `false`
* If enabled, can provide a sample rate value via `tracesSampleRate` property in the option.

This works for my use case, but this is my first contribution to this package so please provide any and all feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,8 @@ Normally, setting required DSN information would be enough.
       vueOptions: {
         tracing: true,
         tracingOptions: {
+          hooks: [ 'mount', 'update' ],
+          timeout: 2000,
           trackComponents: true
         }
       },
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ Normally, setting required DSN information would be enough.
   ```
   - See https://docs.sentry.io/platforms/node/pluggable-integrations/ for more information
 
-### enableTracing
+### tracing
 - Type: `Boolean` or `Object`
   - Default: `false`
   - Enables the BrowserTracing integration for client performance monitoring
--- a/README.md
+++ b/README.md
@@ -329,6 +329,27 @@ Normally, setting required DSN information would be enough.
   ```
   - See https://docs.sentry.io/platforms/node/pluggable-integrations/ for more information
 
+### enableTracing
+- Type: `Boolean` or `Object`
+  - Default: `false`
+  - Enables the BrowserTracing integration
+  - A sampling rate can be provided if passing an object (if not provided, the default rate is 1.0):
+  ```js
+    {
+      tracesSampleRate: 0.4
+    }
+  ```
+  - Note that when enabled, the following is added to the `Vue` integration configuration:
+  ```js
+    {
+      tracing: true,
+      tracingOptions: {
+        trackComponents: true
+      }
+    }
+  ```
+  - See https://docs.sentry.io/platforms/javascript/guides/vue/#monitor-performance for more information
+
 ### config
 - Type: `Object`
   - Default: `{
@@ -354,13 +375,13 @@ Normally, setting required DSN information would be enough.
 - Type: `Object`
   - Default: Refer to `module.js` since defaults include various options that also change dynamically based on other options.
   - Options passed to `@sentry/webpack-plugin`. See documentation at https://github.com/getsentry/sentry-webpack-plugin/blob/master/README.md
-  
+
 ### requestHandlerConfig
 - Type: `Object`
   - Default: `{
   }`
   - Options passed to `requestHandler` in `@sentry/node`. See: https://docs.sentry.io/platforms/node/guides/express/
-  
+
 ## Submitting releases to Sentry
 Support for the [sentry-webpack-plugin](https://github.com/getsentry/sentry-webpack-plugin) was introduced [#a6cd8d3](https://github.com/nuxt-community/sentry-module/commit/a6cd8d3b983b4c6659e985736b19dc771fe7c9ea). This can be used to send releases to Sentry. Use the publishRelease  option to enable this feature.
 
--- a/README.md
+++ b/README.md
@@ -332,23 +332,24 @@ Normally, setting required DSN information would be enough.
 ### enableTracing
 - Type: `Boolean` or `Object`
   - Default: `false`
-  - Enables the BrowserTracing integration
-  - A sampling rate can be provided if passing an object (if not provided, the default rate is 1.0):
+  - Enables the BrowserTracing integration for client performance monitoring
+  - Takes the following object configuration format (default values shown):
   ```js
     {
-      tracesSampleRate: 0.4
+      tracesSampleRate: 1.0,
+      vueOptions: {
+        tracing: true,
+        tracingOptions: {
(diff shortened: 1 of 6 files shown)


speery4 (approved):

lib/core/hooks.js
```diff
@@ -83,6 +83,9 @@ export async function buildHook (moduleContainer, options, logger) {
       }
     }
   }
+  if (options.enableTracing) {
+    options.clientConfig.tracesSampleRate = options.enableTracing.tracesSampleRate ? options.enableTracing.tracesSampleRate : 1
```
rchl, line 87: I suppose it can be just:
```suggestion
    options.clientConfig.tracesSampleRate = options.enableTracing.tracesSampleRate || 1
```

README.md
```diff
@@ -329,6 +329,27 @@ Normally, setting required DSN information would be enough.
...
+  - Note that when enabled, the following is added to the `Vue` integration configuration:
+  ```js
+    {
+      tracing: true,
+      tracingOptions: {
+        trackComponents: true
+      }
+    }
+  ```
+  - See https://docs.sentry.io/platforms/javascript/guides/vue/#monitor-performance for more information
```
rchl, line 351: First of all, great work. It's almost perfect. I'd just like to have it possible to be able to override the vue integration options also.

The structure of this default configuration object could look like so:

```
{
    tracesSampleRate: 1,
    vueOptions: {
        tracing: true,
        tracingOptions: {
            trackComponents: true
        }
    },
}
```
Then whatever user passes will be merged with those default options using `deepMerge`.

I'm also thinking that [server-side tracing](https://docs.sentry.io/platforms/node/performance/) should be supported. If not now then in the future.
So I'm wondering whether there should be both `enableServerTracing` and `enableClientTracing` instead of just `enableTracing`. But I can't make up my mind about that... I guess we could instead add `disableServerTracing` within this configuration object in the future, or something.

rchl (changes requested):

README.md
```diff
@@ -329,6 +329,27 @@ Normally, setting required DSN information would be enough.
   ```
   - See https://docs.sentry.io/platforms/node/pluggable-integrations/ for more information
 
+### enableTracing
+- Type: `Boolean` or `Object`
+  - Default: `false`
+  - Enables the BrowserTracing integration
+  - A sampling rate can be provided if passing an object (if not provided, the default rate is 1.0):
```
rchl, line 336: I feel that the default sampling rate should be lower (maybe 0.6 or even 0.4?). If users want to be overwhelmed they can raise it to 1. What do you think?

README.md
```diff
@@ -329,6 +329,27 @@ Normally, setting required DSN information would be enough.
   ```
   - See https://docs.sentry.io/platforms/node/pluggable-integrations/ for more information
 
+### enableTracing
```
rchl, line 332: We should probably mention that `@sentry/tracing` should be installed manually when using this option.

I guess it should work regardless because `@sentry/node` seems to depend on it but probably better to be more explicit.

README.md
rchl: And `BrowserTracing` integration also takes various options so it should be possible to pass those.

README.md
jonthemiller: My thought was that a new user testing out the feature (just setting the option to `true`) would want their first request captured, with no additional configuration. The Sentry docs do recommend changing this to a lower value, just not sure what a sane value would be. Can always heavily encourage users to configure this key.

rchl: OK, let's have a warning/advice similar to what Sentry says.

README.md
jonthemiller: Agreed that server side would be a good add, just didn't want to tackle it as part of this PR currently.

jonthemiller:
Would you prefer this be named `enableTracing` or just `tracing`?

rchl:
> Would you prefer this be named `enableTracing` or just `tracing`?

Since there is no clear convention already, I think `tracing` would be enough.

jonthemiller:
I think I addressed everything - let me know if anything else needs changed.

lib/core/hooks.js
```diff
@@ -83,6 +83,19 @@ export async function buildHook (moduleContainer, options, logger) {
...
     }
   }
+  if (options.tracing) {
+    options.tracing = deepMerge.all([{
+      tracesSampleRate: 1.0,
+      vueOptions: {
+        tracing: true,
+        tracingOptions: {
+          trackComponents: true
+        }
```
rchl, line 93: Let's add all those (plus types):

```ts
/** Vue specific configuration for Tracing Integration  */
interface TracingOptions {
    /**
     * Decides whether to track components by hooking into its lifecycle methods.
     * Can be either set to `boolean` to enable/disable tracking for all of them.
     * Or to an array of specific component names (case-sensitive).
     */
    trackComponents: boolean | string[];
    /** How long to wait until the tracked root activity is marked as finished and sent of to Sentry */
    timeout: number;
    /**
     * List of hooks to keep track of during component lifecycle.
     * Available hooks: 'activate' | 'create' | 'destroy' | 'mount' | 'update'
     * Based on https://vuejs.org/v2/api/#Options-Lifecycle-Hooks
     */
    hooks: Operation[];
}
declare type Operation = 'activate' | 'create' | 'destroy' | 'mount' | 'update';
```
Unfortunately, we can't import types directly from sentry so have to duplicate

jonthemiller: What should be used for the defaults for `timeout` and `hooks`? Looking at Sentry's source, timeout is `2000` and hooks are `mount` and `update` ?

https://github.com/getsentry/sentry-javascript/blob/master/packages/integrations/src/vue.ts#L164-L169

rchl: Yeah. Let's use the same.

As much as I hate duplicating stuff and risking things getting out of sync, it makes it quite more transparent for the user in case he wants to modify some of those.

types/sentry.d.ts
```diff
@@ -14,6 +14,17 @@ export interface LazyConfiguration {
...
 
+export interface TracingConfiguration {
+    tracesSampleRate?: number
+    vueOptions?: {
+        tracing?: boolean
+        tracingOptions?: {
+            trackComponents?: boolean
+        }
+    },
+    browserOptions?: unknown
```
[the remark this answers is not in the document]
jonthemiller: added

types/sentry.d.ts
```diff
@@ -14,6 +14,17 @@ export interface LazyConfiguration {
...
 }
 
+export interface TracingConfiguration {
+    tracesSampleRate?: number
+    vueOptions?: {
+        tracing?: boolean
+        tracingOptions?: {
+            trackComponents?: boolean
+        }
+    },
```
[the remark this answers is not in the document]
jonthemiller: fixed

rchl (approved):

rchl:
Thank you.

jonthemiller:
Thank you as well.

PR #254, opened 2020-11-26
feat: Add runtime config support

### LINKED ISSUE #202: Plugin not using runtime variables from $config

julien1619:
This PR adds a new configuration key: `publicRuntimeConfigKey`, it defaults to `sentry` when publicRuntimeConfig exists.

It fixes #202, and inspired by #203, thanks to @Ledarath


lib/plugin.client.js
```diff
@@ -10,7 +10,13 @@ import { Integrations as TracingIntegrations } from '@sentry/tracing'
 export default function (ctx, inject) {
   <% if (options.initialize) { %>
   /* eslint-disable object-curly-spacing, quote-props, quotes, key-spacing, comma-spacing */
-  const config = <%= serialize(options.config) %>
+  const config = <%= JSON.stringify(options.config) %>
```
rchl, line 13: What's the reason for this change?

lib/plugin.client.js
```diff
@@ -10,7 +10,13 @@ import { Integrations as TracingIntegrations } from '@sentry/tracing'
...
+
+  const publicRuntimeConfigKey = <%= serialize(options.publicRuntimeConfigKey) %>
+  if (ctx.$config && publicRuntimeConfigKey) {
+    Object.assign(config, ctx.$config[publicRuntimeConfigKey])
```
rchl, line 17: `Object.assign` is a very poor way of merging options. Should use `lodash.merge` instead.

(Currently this module uses `deepmerge` for merging options but we shouldn't use that one on the client so we can replace `deepmerge` with `lodash.merge` altogether.)

rchl (changes requested):
Missing:
 - documentation
 - similar change in `plugin.lazy.js`
 - What about server-side? Even though on server-side options are "runtime", the user might want to use the runtime config for both server and client-side rather than having to specify them in both places.

lib/plugin.client.js
julien1619: No reason at all, it's a mistake due to the rebase. I'll fix it.

lib/plugin.client.js
julien1619: Ok, I'll do the replacement.

julien1619:
Thanks for the review, I'll do that tomorrow.

julien1619:
@rchl I did all the requested changes.

- Add documentation
- Replace `deepmerge` and all `Object.assign` by `lodash.merge`
- Add Runtime Config to lazy mode and servier-side
- Restore `serialize` instead of `JSON.stringify`

julien1619:
I've tested it on my site and there is a bug in the config merge, I'll investigate tomorrow. For now I'll set the draft mode on this Pull Request.

julien1619:
It works now. @rchl can you review it again? Thanks!

julien1619:
@rchl Just did a rebase onto master. Is there something blocking this PR from being merged? Thanks!

rchl:
Just lack of time but will try to handle soon

lib/core/hooks.js
```diff
@@ -1,5 +1,5 @@
 import { resolve, posix } from 'path'
-import deepMerge from 'deepmerge'
+import merge from 'lodash.merge'
```
rchl, line 2: Sorry to go back and forth on this but for the size reasons we've recently switched to `klona` in `nuxt-i18n` (https://github.com/nuxt-community/i18n-module/pull/981) and I'd like to do the same here.

README.md
```diff
@@ -374,6 +388,12 @@ Normally, setting required DSN information would be enough.
   }`
   - Specified keys will override common Sentry options for client sentry plugin
 
+### publicRuntimeConfigKey
+- Type: `String`
+  - Default: If Nuxt >=2.13 `sentry` , else `undefined`
+  - Specified object in Nuxt config in `publicRuntimeConfig[publicRuntimeConfigKey]` will override config at runtime. See documentation at https://fr.nuxtjs.org/docs/2.x/configuration-glossary/configuration-runtime-config/
```
rchl, line 394: Please link using canonical URL https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-runtime-config/
The current one is french.

lib/module.js
```diff
@@ -14,6 +14,7 @@ export default function SentryModule (moduleOptions) {
     dsn: process.env.SENTRY_DSN || '',
     disabled: envToBool(process.env.SENTRY_DISABLED) || false,
     initialize: envToBool(process.env.SENTRY_INITIALIZE) || true,
+    publicRuntimeConfigKey: this.options.publicRuntimeConfig ? 'sentry' : undefined,
```
rchl, line 17: Can we always fallback to `'sentry'`? The current way is wrong type-wise since you are potentially assigning `undefined` to a required type.

lib/plugin.client.js
```diff
@@ -1,4 +1,5 @@
 import VueLib from 'vue'
+import merge from 'lodash.merge'
```
rchl, line 2: Maybe let's import it (or `klona` rather) dynamically from the place where it's needed since it won't always be needed.

README.md
```diff
@@ -91,6 +91,20 @@ Set `lazy: true` in your module options to load Sentry lazily on the client. Thi
 
 You can also pass a lazy config object in your module options (see [options](#lazy) for more information).
 
+## Runtime Configuration
+
+Define a configuration object at runtime in Nuxt config `publicRuntimeConfig[]` to set the environment at runtime for example.
```
rchl, line 96: - It appears that this runtime configuration is only for setting/overriding the options provided in the `config` option. So the one that is passed to `Sentry.init()`. This is not clear at all. The user might think she/he can specify all module options here (which is how it should really work IMO)
- This section could be moved closer to the `Options` section. Probably even be inside of it.
- If making runtime configuration work for "all" options, we should annotate which option is actually possible to override with runtime config since not all options really can. For example, `lazy` can only be resolved during build-time. Same for `disableClientSide`, `disableServerSide` and probably some other.

rchl (changes requested):

lib/module.js
julien1619: Good catch! I'll change it.

README.md
julien1619: I agree that it should be better specified in documentation.

I didn't want to allow anything else than in the `config` option to be set at runtime because at compile-time this module is doing some work to declare a release, upload the source map, etc. So only a limited set of options can be changed at runtime without breaking the consistency of the configuration, so I think keeping it limited to the `config` option make it clearer.

rchl: I'm fine if your PR only adds support for overriding the `config` but we can't do it the way you did here as it's not future proof in case we would want to extend it later to support other options.

Also, that solution doesn't support overriding `config` through `clientConfig` or `serverConfig`.

So please change it so that user is required to use this structure:

```js
publicRuntimeConfig: {
  sentry: {
    config: {
      // ...
    }
  }
}
```

You can just support `config`, `clientConfig`, and `serverConfig` and I can later do a follow-up PR to try to support other options.

julien1619: Ok I understand, I'll fix it.

lib/core/hooks.js
julien1619: In fact, I think that Klona is for cloning, not merging. We need to merge multiple objects, so I don't think I can replace `lodash.merge` with `klona`.

lib/plugin.client.js
rchl: That's true, sorry for creating confusion.

julien1619:
@rchl We did all the changes requested. The `publicRuntimeConfig` can now be used to override `config`, `clientConfig` and `serverConfig` (Thanks to @Ledarath ) Can you review it? Thanks!

rchl:
Would you be able to enable the option that allows maintainers to push to this branch? I could do some minor changes myself instead of the long way with comments and turnarounds.

julien1619:
Done! I invited you directly, the option "Allow edits from maintainers" wasn't available.

rchl:
OK, I've done some changes. You can have a look through the commits I've added and check if it looks good to you.

julien1619:
LGTM 👍

rchl (approved):

rchl:
Thanks!

samturrell:
Sorry to jump in on this late. To confirm, the `dsn` still has to be provided at build time and will be baked into the built code? There's no way of providing this as a runtime config value?

rchl:
A bit weird maybe but you can probably set any `dsn` in build-time settings and override it through runtime settings.

PR #259, opened 2020-12-13
docs: add nuxt content documentation

### LINKED ISSUE #255: Documentation
Hi @rchl 👋 

Do you think it could be a great idea to purpose a documentation for `sentry-module` ? 
If yes, i can purpose a MR in this way to have a unified documentation with all others nuxt module through nuxt-content. 

What do you think ? 😀

Geminii:
Creation of a documentation using nuxt-content to uniformize all nuxt modules documentation ❤️

- [ ] define url for documentation online (README / settings.json) 
- [ ] icon
- [ ] preview light
- [ ] preview dark
- [ ] logo light
- [ ] logo dark

Resolves #255


README.md
```diff
@@ -9,388 +11,17 @@
...
+<p align="center">
+  <a href="https://sentry.nuxtjs.org">Read Documentation</a>
+</p>
+
 ## Features
 
 The module enables error logging through [Sentry](http://sentry.io).
 
 - **Please note** that version 2.2.0 of this package removed the older `public_key` and `private_key` options, since the updated Sentry packages don't support these anymore.
 - **Please note** that version 2.0.0 of this package introduces a breaking change. See [#30](https://github.com/nuxt-community/sentry-module/pull/30) for more information.
```
rchl, line 23: Let's just remove those notes now. Both from here and documentation. Those are very outdated.

docs/content/en/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+  Please note that version 2.2.0 of this package removed the older `public_key` and `private_key` options, since the updated Sentry packages don't support these anymore.
+
+</alert>
+
+<alert type="warning">
+
+  Please note that version 2.0.0 of this package introduces a breaking change. See [#30](https://github.com/nuxt-community/sentry-module/pull/30) for more information.
+
+</alert>
```
rchl, line 27: Let's remove those.

docs/content/en/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+## More Resources
+
+* [GitHub](https://github.com/nuxt-community/sentry-module)
+* [Releases](https://github.com/nuxt-community/sentry-module/releases)
+* [MIT Licence](./LICENSE)
```
rchl, line 33: This link would have to be absolute.

docs/content/en/guide/setup.md
```diff
@@ -0,0 +1,49 @@
...
+  ```bash
+  npm install @nuxtjs/sentry
+  ```
+
+  </code-block>
+</code-group>
+
+Then, add `@nuxtjs/sentry` to the `modules` section of `nuxt.config.js`:
+
+```js[nuxt.config.js]
```
rchl, line 35: Please add a space after the initial `js`. Otherwise, it breaks syntax highlighting in some editors.

docs/content/en/sentry/lazy-loading.md
```diff
@@ -0,0 +1,63 @@
...
+You can also pass a lazy config object in your module options (see [options](/sentry/options#lazy) for more information).
+
+### Injected properties
+
+#### `$sentry` (mocked)
+- Type: `Object`
+
+Normally `$sentry` would always refer to the `@sentry/browser` API. But if we lazy load Sentry this API wont be available until Sentry has loaded. If you don't want to worry about whether Sentry is loaded or not, a mocked Sentry API is injected into the Nuxt.js context that will execute all Sentry API calls once Sentry is loaded
+
+See: [`injectMock`](/sentry/options#lazy) and [`mockApiMethods`](/sentry/options#lazy) options below
```
rchl, line 25: Let's remove "below".

docs/content/en/sentry/options.md
```diff
@@ -0,0 +1,251 @@
...
+      webpackPreload: false
+    }
+  ```
+    - **injectMock**
+      - Type: `Boolean`
+        - Default: `true`
+        - Whether a Sentry mock needs to be injected that captures any calls to `$sentry` API methods while Sentry has not yet loaded. Captured API method calls are executed once Sentry is loaded
+      > When `injectMock: true` this module will also add a window.onerror listener. If errors are captured before Sentry has loaded then these will be reported once Sentry has loaded using sentry.captureException
+        ```js
+        // pages/index.vue
```
rchl, line 37: Could use the `[filepath]` syntax for file name.

docs/content/en/sentry/options.md
```diff
@@ -0,0 +1,251 @@
...
+          this.$sentry.captureMessage('Hello!')
+        },
+        ```
+
+    - **injectLoadHook**
+      - Type: `Boolean`
+        - Default: `false`
+        - By default Sentry will be lazy loaded once `window.onNuxtReady` is called. If you want to explicitly control when Sentry will be loaded you can set `injectLoadHook: true`. The module will inject a `$sentryLoad` method into the Nuxt.js context which you need to call once you are ready to load Sentry
+        ```js
+        // layouts/default.vue
```
rchl, line 52: Could use the `[filepath]` syntax for file name.

docs/content/en/sentry/options.md
```diff
@@ -0,0 +1,251 @@
...
+        ```js
+        // nuxt.config.js
+        sentry: {
+          lazy: {
+            mockApiMethods: ['captureMessage']
+          }
+        }
+
+        // pages/index.vue
+        mounted() {
```
rchl, line 77: Could use the `[filepath]` syntax for file name and maybe split into two code blocks.

docs/content/en/sentry/options.md
```diff
@@ -0,0 +1,251 @@
...
+          trackComponents: true
+        }
+      },
+      browserOptions: {}
+    }
+  ```
+  - Sentry docs strongly recommend reducing the `tracesSampleRate` value; it should be between 0.0 and 1.0 (percentage of requests to capture)
+  - The `vueOptions` are passed to the `Vue` integration, see https://docs.sentry.io/platforms/javascript/guides/vue/#monitor-performance for more information
+  - `browserOptions` are passed to the `BrowserTracing` integration, see https://github.com/getsentry/sentry-javascript/tree/master/packages/tracing for more information
+  - `@sentry/tracing` should be installed manually when using this option (it is currently a dependency of `@sentry/node`)
```
rchl, line 219: Can we move it before the "Type" and make it an `<alert>` element? I feel like this should be more exposed.

rchl (changes requested):

Geminii:
@rchl I take in consideration all of your feedbacks and update static ressources

docs/nuxt.config.js
```diff
@@ -1,3 +1,7 @@
 import theme from '@nuxt/content-theme-docs'
 
-export default theme({})
+export default theme({
+  docs: {
+    primaryColor: '#564995'
```
rchl, line 5: This color is very bad in terms of contrast when using the dark theme.

Screenshot 2020-12-14 at 23 23 21

(The contrast can be tested with the Lighthouse developer tool, using accessibility checks)

Geminii: Thanks for the tip about "Lighthouse". i never used it, very cool to discover it :) 
I will improve the primary color so 👍

rchl: Is it possible to define different colors depending on dark/light theme? I'm asking because with latest changes now the light theme is not contrasting enough (not that some of those are issues from Nuxt theme itself but all the purple links also have a contrast problem):

Screenshot 2020-12-16 at 09 17 03

Atinux:
Could you add a description for each markdown file too in the frontmatter @Geminii ?

Geminii:
Indeed, i will add it to improve SEO performance 😍

docs/nuxt.config.js
Geminii: Currently no unfortunatly. We can only define one color for both ... 
Later with new docs theme, we we'll be able to 😀 
Can you purpose a doc color to solve this contrast for light & dark theme around the logo color ?

rchl: > Can you purpose a doc color to solve this contrast for light & dark theme around the logo color ?

That's not possible, I think. We need to be able to set different colors for dark/light mode.

rchl:
I think one remaining thing is to have a favicon. Currently, it's using the generic Nuxt Content one.

The contrast issue can probably be handled separately as it's not an issue with those changes per-se.

Geminii:
> I think one remaining thing is to have a favicon. Currently, it's using the generic Nuxt Content one.

There is a favicon for Sentry available here : https://github.com/nuxt-community/sentry-module/blob/ff795b42eb19852ab5766debff5b51fb36d92461/docs/static/icon.png
Define an `icon.png` will override the default nuxt-content favicon :) 
You can get more information about it : https://content.nuxtjs.org/themes/docs/#static

> The contrast issue can probably be handled separately as it's not an issue with those changes per-se.

As you wish, else we can wait for new doc theme but maybe not available before few weeks ^^

rchl:
> There is a favicon for Sentry available here : [`ff795b4`/docs/static/icon.png](https://github.com/nuxt-community/sentry-module/blob/ff795b42eb19852ab5766debff5b51fb36d92461/docs/static/icon.png)

Yes, but it doesn't seem to work for me with `yarn dev`. I still see the default nuxt content icon in the tab.

PR #275, opened 2021-02-16
fix: support configuring "CaptureConsole" and "Debug" integrations

### LINKED ISSUE #274: How can I add pluggable integrations not listed in SENTRY_BROWSER_INTEGRATIONS to sentry module?
I would like to add e.g. CaptureConsole integration. But if I do it in the nuxt.config.js, I'll get 

`Sentry integration "CaptureConsole" is not recognized and will be ignored. nuxt:sentry`

because the integrations is missing in array [SENTRY_BROWSER_INTEGRATIONS](https://github.com/nuxt-community/sentry-module/blob/master/lib/core/hooks.js#L10).

This is my part of nuxt config:

```
  sentry: {
    dsn: ``,
    config: {
      maxBreadcrumbs: 100,
      // ...
    },
    clientIntegrations: {
      CaptureConsole: {
        levels: ['error'],
      },
    },
  },
```

Maybe it'll help, if you add information about these integrations to doc https://sentry.nuxtjs.org/sentry/options#clientintegrations.

- Caused by: https://github.com/nuxt-community/sentry-module/pull/257
- Related issue: https://github.com/nuxt-community/sentry-module/issues/191?

Thx

rchl:
Resolves #274

### DIFF
--- a/lib/core/hooks.js
+++ b/lib/core/hooks.js
@@ -6,7 +6,7 @@ import WebpackPlugin from '@sentry/webpack-plugin'
 import { canInitialize, clientSentryEnabled, serverSentryEnabled } from './utils'
 
 const SERVER_CONFIG_FILENAME = 'sentry.server.config.js'
-const SENTRY_DEFAULT_INTEGRATIONS = ['CaptureConsole', 'Debug', 'Dedupe', 'ExtraErrorData', 'ReportingObserver', 'RewriteFrames', 'Vue']
+const SENTRY_PLUGGABLE_INTEGRATIONS = ['CaptureConsole', 'Debug', 'Dedupe', 'ExtraErrorData', 'ReportingObserver', 'RewriteFrames', 'Vue']
 const SENTRY_BROWSER_INTEGRATIONS = ['InboundFilters', 'FunctionToString', 'TryCatch', 'Breadcrumbs', 'GlobalHandlers', 'LinkedErrors', 'UserAgent']
 const SENTRY_SERVER_INTEGRATIONS = ['CaptureConsole', 'Debug', 'Dedupe', 'ExtraErrorData', 'RewriteFrames', 'Modules', 'Transaction']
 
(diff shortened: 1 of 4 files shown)


lib/core/hooks.js
```diff
@@ -6,8 +6,9 @@ import WebpackPlugin from '@sentry/webpack-plugin'
 import { canInitialize, clientSentryEnabled, serverSentryEnabled } from './utils'
 
 const SERVER_CONFIG_FILENAME = 'sentry.server.config.js'
-const SENTRY_DEFAULT_INTEGRATIONS = ['Dedupe', 'ExtraErrorData', 'ReportingObserver', 'RewriteFrames', 'Vue']
+const SENTRY_DEFAULT_INTEGRATIONS = ['CaptureConsole', 'Debug', 'Dedupe', 'ExtraErrorData', 'ReportingObserver', 'RewriteFrames', 'Vue']
```
JkmAS, line 9: I am not sure about moving `CaptureConsole` & `Debug` to the default integrations (https://docs.sentry.io/platforms/javascript/configuration/integrations/default/). `SENTRY_BROWSER_INTEGRATIONS` var makes more sense to me. 

But I don't know the difference between these two arrays (`SENTRY_BROWSER_INTEGRATIONS` vs `SENTRY_BROWSER_INTEGRATIONS`)

JkmAS (commented):
I have a one question :point_up: . Otherwise LGTM, thx.

lib/core/hooks.js
rchl: The meaning of DEFAULT vs BROWSER is that the DEFAULT maps to `@sentry/integrations` import while BROWSER to `@sentry/browser` import. So it's correct as is.

I admit it's confusing when seen in the context of Sentry documentation but this is just internal naming that users won't see. I can rename DEFAULT to PLUGGABLE I guess.

JkmAS: I see. 

> I can rename DEFAULT to PLUGGABLE I guess.

good idea, thx

PR #276, opened 2021-02-17
feat: support for registering external integrations

### LINKED ISSUE #240: Allow plugins from outside @sentry/integrations
### Is your feature request related to a problem? Please describe.
I'd like to use the [Sentry rrweb plugin](https://docs.sentry.io/platforms/javascript/configuration/integrations/rrweb/) to have session replays with some of my errors. However, it seems I can only add integrations that are part of `@sentry/integrations` because of the way they are imported and used [here](https://github.com/nuxt-community/sentry-module/blob/master/lib/plugin.client.js#L4).

### Describe the solution you'd like
The ability to declare plugins from outside of `@sentry/integrations` (eg. `import SentryRRWeb from "@sentry/rrweb"`). I'm not sure what the best way would be to alter the current configuration in `nuxt.config.js` to facilitate this. 

### Describe alternatives you've considered
I don't know any other ways to solve this at the moment.

rchl:
Resolves #240


rchl:
Actually, there will likely have to be separate options for server/client as the server doesn't run in webpack context so doesn't have access to Nuxt context and doesn't support aliases in the path...

rchl:
Was planning to add tests for this feature but various sentry-testkit [bugs](https://github.com/wix/sentry-testkit/issues?q=is%3Aopen+is%3Aissue+author%3Archl+sort%3Aupdated-desc+) made it impossible.

mammuth:
Awesome, thanks! :100: (I'll try configuring fullstory with this)

Btw, will you also publish a new package version for this commit?

rchl:
I've released it as a beta for now as there are breaking changes: https://github.com/nuxt-community/sentry-module/releases/tag/v6.0.0-beta.4

Issue #285, opened 2021-03-18
Lazy loading and Runtime Config


mikeapr4:
### Version
@nuxtjs/sentry: 5.0.2
nuxt: 2.15.3

### Sentry configuration
```
  sentry: {
    dsn: "https://public@sentry.example.com/1",
    lazy: true,
  },

  publicRuntimeConfig: {
    sentry: {
      environment: "test"
    }
  }
```

### Reproduction Link
https://codesandbox.io/s/bold-einstein-l2fq4
(note the console output inside the codesandbox doesn't seem to show the error, you need to open the preview URL and look at the console: https://l2fq4.sse.codesandbox.io/ )

### Steps to reproduce
Load a page (initializing Sentry)

### What is Expected?
Config to be merged without problems

### What is actually happening?
`TypeError: merge is not a function`

rchl:
Fixed with #286

rchl:
Released fix in https://github.com/nuxt-community/sentry-module/releases/tag/v5.0.3

PR #286, opened 2021-03-18
fix: crash on lazy loading when using Runtime Config

mikeapr4:
Solution for https://github.com/nuxt-community/sentry-module/issues/285

The error occurs because we do a dynamic import of `lodash.merge` which is a CommonJS module.


rchl:
Thanks. I've added a test also.

PR #288, opened 2021-03-22
fix(docs): fixes typography & content styling

aldy505:
This is what currenty [Sentry docs](https://sentry.nuxtjs.org/) shows:

Current docs

It turns out, on `@nuxt/content-theme-docs` `^0.10.0` and higher, the styling's broken. Reverting the version back to `^0.9.0` fixes the problem.

New docs

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -9,7 +9,7 @@
     "generate": "nuxt generate"
   },
   "dependencies": {
-    "@nuxt/content-theme-docs": "^0.10.2",
+    "@nuxt/content-theme-docs": "^0.9.0",
     "nuxt": "^2.15.3"
   }
 }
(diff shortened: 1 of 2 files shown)


rchl:
Thanks!

rchl:
Thank didn't really fix anything, did it?

I think it's more about the sub dependencies of that package rather than the package version itself.

aldy505:
It fixed the problem on my machine, installing nuxt/content v0.10.0 brokes the docs again on mine, I've tested it.

rchl:
It's still broken on https://sentry.nuxtjs.org/
I'm fixing it with #291

As I said, it's not necessarily the version of nuxt/content that matters but one of the dependencies referenced in the lock file. So the nuxt/context version can be higher as long as that dependency is not updated. Not sure which one though.

PR #302, opened 2021-04-04
fix: handle commonjs/es import styles for @sentry/cli

### LINKED ISSUE #301: Automatic config.release from @sentry/cli fails because of undefined import
Hi, I found that `SentryCli` is `undefined` when importing on this line:

https://github.com/nuxt-community/sentry-module/blob/eed66dbc820678a89ba1e874adb2df7956af8f88/lib/core/hooks.js#L43

It works as expected when removing `.default`

```js
 const SentryCli = await import('@sentry/cli')
```

I've tested this fix on `nuxt 2.15.4` on a minimal create-nuxt-app repro: [snake575/nuxt-sentry](https://github.com/snake575/nuxt-sentry)

snake575:
Fixes #301

### DIFF
--- a/lib/core/hooks.js
+++ b/lib/core/hooks.js
@@ -40,7 +40,7 @@ export async function buildHook (moduleContainer, options, logger) {
     // Determine "config.release" automatically from local repo if not provided.
     try {
       // @ts-ignore
-      const SentryCli = await import('@sentry/cli')
+      const SentryCli = await (import('@sentry/cli').then(m => m.default || m))
       const cli = new SentryCli()
       options.config.release = (await cli.releases.proposeVersion()).trim()
     } catch {
--- a/lib/core/hooks.js
+++ b/lib/core/hooks.js
@@ -40,7 +40,7 @@ export async function buildHook (moduleContainer, options, logger) {
     // Determine "config.release" automatically from local repo if not provided.
     try {
       // @ts-ignore
-      const SentryCli = (await import('@sentry/cli')).default
+      const SentryCli = await import('@sentry/cli')
       const cli = new SentryCli()
       options.config.release = (await cli.releases.proposeVersion()).trim()
     } catch {


lib/core/hooks.js
```diff
@@ -40,7 +40,7 @@ export async function buildHook (moduleContainer, options, logger) {
     // Determine "config.release" automatically from local repo if not provided.
     try {
       // @ts-ignore
-      const SentryCli = (await import('@sentry/cli')).default
+      const SentryCli = await import('@sentry/cli')
```
rchl, line 43: Try something like this instead:
```suggestion
      const SentryCli = await (import('@sentry/cli').then(m => m.default || m))
```
Whether the import contains the `default` export can depend on the user setting so the code should handle both cases.

rchl (changes requested):

lib/core/hooks.js
snake575: Thanks! that's a better option

snake575:
@rchl I noticed that `@sentry/webpack-plugin` already gets the release version with `proposeVersion` by default:

https://github.com/getsentry/sentry-webpack-plugin/blob/fbd610c22abf94b41f3056d7f69fb32a96e95314/src/index.js#L86

Maybe it's not necessary to implement the same feature on this module?

rchl:
> @rchl I noticed that `@sentry/webpack-plugin` already gets the release version with `proposeVersion` by default:
> 
> [getsentry/sentry-webpack-plugin@`fbd610c`/src/index.js#L86](https://github.com/getsentry/sentry-webpack-plugin/blob/fbd610c22abf94b41f3056d7f69fb32a96e95314/src/index.js#L86)
> 
> Maybe it's not necessary to implement the same feature on this module?

I think the `release` needs to be set in the config created at build-time, even if you have `publishRelease` is enabled as webpack-plugin won't update the module's config.

Feel free to correct me if I'm wrong as I don't fully remember how everything works now.

snake575:
> I think the `release` needs to be set in the config that is generated at build-time even if `publishRelease` is enabled as webpack-plugin won't update the module's config.
> 
> Feel free to correct me if I'm wrong as I don't fully remember how everything works now.

I made new branch deleting the implementation to test this (it works). Maybe was implemented to fix some edge case?

https://github.com/nuxt-community/sentry-module/compare/master...snake575:sentry-webpack-default-release

The uploaded `release` matches this branch last commit ID

https://github.com/snake575/nuxt-sentry/tree/sentry-webpack-default-release

https://github.com/snake575/nuxt-sentry/commit/2ae0ea493bf09fc219661f77c535b88eca2b5e1c

image

rchl:
But are issues reported with the correct version still?

Besides, I think that even when you have not enabled `publishRelease`, you still want the correct version to be reported in the issues.

snake575:
You're right, on the branch eliminating the implementation, issues get reported on sentry but not on any release.

PR #317, opened 2021-04-29
docs: Page on how to modify report before sending

MentalGear:
Added Page on how to use hook to modify reports.

### DIFF
--- a/docs/content/en/sentry/hooks
+++ b/docs/content/en/sentry/hooks
@@ -1,0 +1,24 @@
+---
+title: Report Pre-processing (Hooks)
+description: 'Modify data before transmission'
+position: 6
+category: Sentry
+---
+
+If you would like to modify a report's captured data before transmission (e.g. add additional or scrub sensitive data), you can use a [Sentry hook](https://docs.sentry.io/platforms/javascript/configuration/options/#hooks) within your 'config'.
+
+Example of using a hook to trigger [Sentry's native User Feedback/Error Reporting Dialog](https://docs.sentry.io/platforms/javascript/guides/vue/enriching-events/user-feedback/) to gather additional information from the user:
+```
+    config: {
+      environment: process.env.NODE_ENV,
+      beforeSend(event, hint) {
+        // Check if it is an exception, and if so, show the report dialog
+        if (event.exception) {
+          // Native Sentry JS Bug/Feedback Reporter
+          // https://docs.sentry.io/platforms/javascript/guides/vue/enriching-events/user-feedback/
+          Sentry.showReportDialog({ eventId: event.event_id })
+        }
+        return event
+      },
+    },
+```


docs/content/en/sentry/hooks
```diff
@@ -0,0 +1,24 @@
+---
+title: Report Pre-processing (Hooks)
+description: 'Modify data before transmission'
+position: 6
```
rchl, line 4: There is already a page with position 6.
Maybe make it 7 and move page 7 to 8.

docs/content/en/sentry/hooks
```diff
@@ -0,0 +1,24 @@
+---
+title: Report Pre-processing (Hooks)
```
rchl, line 2: I'd prefer non-titleCase as this casing looks a bit random.

`Report pre-processing (hooks)`

docs/content/en/sentry/hooks
```diff
@@ -0,0 +1,24 @@
...
+description: 'Modify data before transmission'
+position: 6
+category: Sentry
+---
+
+If you would like to modify a report's captured data before transmission (e.g. add additional or scrub sensitive data), you can use a [Sentry hook](https://docs.sentry.io/platforms/javascript/configuration/options/#hooks) within your 'config'.
+
+Example of using a hook to trigger [Sentry's native User Feedback/Error Reporting Dialog](https://docs.sentry.io/platforms/javascript/guides/vue/enriching-events/user-feedback/) to gather additional information from the user:
+```
```
rchl, line 11: You can add a language tag (`js`) to get syntax highlighting.

docs/content/en/sentry/hooks
```diff
@@ -0,0 +1,24 @@
...
+    config: {
+      environment: process.env.NODE_ENV,
```
rchl, line 13: Let's remove this line since it's not really relevant.

On a related note, I'm not sure if it's a good idea to control the sentry environment with **that** variable.  I'm thinking that various libraries might check it and enable more optimized/less chatty behavior for `production` and I would like my `production` and `stage` environments both to have as close behavior as possible.

So I would go for a custom ENV variable for deciding the Sentry environment.

docs/content/en/sentry/hooks
```diff
@@ -0,0 +1,24 @@
...
+      beforeSend(event, hint) {
+        // Check if it is an exception, and if so, show the report dialog
+        if (event.exception) {
+          // Native Sentry JS Bug/Feedback Reporter
+          // https://docs.sentry.io/platforms/javascript/guides/vue/enriching-events/user-feedback/
+          Sentry.showReportDialog({ eventId: event.event_id })
+        }
+        return event
+      },
```
rchl, line 22: I don't think this would even work as:
 1. this will also run on the server where it doesn't make sense to show report dialog and that method does likely not even exist there.
 2. The `Sentry` object is not in scope of that function, even if you have it imported in your nuxt config. That's because that function is serialized and loaded separately from a plugin so nothing will be in scope.

For 1 you can add a `process.client` check, or define this hook in the `clientConfig` instead.
For 2, I'm not sure... I can't think of a way to access the Sentry object from here. Unless maybe there is a custom client-side plugin added that exposes sentry on the `window` object.

rchl (changes requested):

rchl:
I've added https://sentry.nuxtjs.org/guide/user-feedback page to documentation so it should cover this case more or less.

PR #325, opened 2021-06-03
docs: clarified publishRelease config requirements

MentalGear:
At least until options for ORG, AUTH and maybe PROJECT are available.

### DIFF
--- a/docs/content/en/sentry/options.md
+++ b/docs/content/en/sentry/options.md
@@ -157,7 +157,7 @@ Normally, just setting DSN would be enough.
 
 - Type: `Boolean`
 - Default: `process.env.SENTRY_PUBLISH_RELEASE || false`
-- this option requires `SENTRY_ORG` and `SENTRY_AUTH_TOKEN` set via [env/rc/properties file](https://docs.sentry.io/product/cli/configuration/#sentry-cli-working-with-projects)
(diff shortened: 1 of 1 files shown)


docs/content/en/sentry/options.md
```diff
@@ -157,6 +157,7 @@ Normally, just setting DSN would be enough.
 
 - Type: `Boolean`
 - Default: `process.env.SENTRY_PUBLISH_RELEASE || false`
+- this option requires `SENTRY_ORG` and `SENTRY_AUTH_TOKEN` set via [env/rc/properties file](https://docs.sentry.io/product/cli/configuration/#sentry-cli-working-with-projects)
```
rchl, line 160: Finding this a bit **too specific** since it mentions the ENV variables but if you are using the rc file you don't really work with those keys.

I'd suggest a more generic message like:

```suggestion
- This option requires the organization slug, project name and the sentry authentication token to be provided via [environment variables or a properties file](https://docs.sentry.io/product/cli/configuration/#sentry-cli-working-with-projects). So for example when using environment variables you'd set `SENTRY_AUTH_TOKEN`, `SENTRY_ORG` and `SENTRY_PROJECT`.
```

rchl (changes requested):

docs/content/en/sentry/options.md
MentalGear: Actually using `export SENTRY_ORG` also works for the rc file, but you're right with the properties file and the more general version is appreciated.

rchl (approved):

rchl:
Thanks!

PR #373, opened 2021-11-24
docs: clarify that setCommits is not set by default

### LINKED ISSUE #372: Deprecated attachCommits actually still required
[Docs say not to use `attachCommits`](https://sentry.nuxtjs.org/sentry/options#attachcommits) but[ the code only uses the replacement `publishRelease.setCommits` when `attachCommits` is set](https://github.com/nuxt-community/sentry-module/blob/607511a4e4c6000fbea78e811249839d4a1f7272/lib/core/hooks.js#L222).

bmulholland:
Closes #372

### DIFF
--- a/docs/content/en/sentry/options.md
+++ b/docs/content/en/sentry/options.md
@@ -174,7 +174,7 @@ sentry: {
     project: 'my-project',
     // Attach commits to the release (requires that the build triggered within a git repository).
     setCommits: {
-      auto: true  // set by default
+      auto: true
     }
   }
 }


bmulholland:
Hold this - I'll fix the commit and add a note to be careful with nuxt/emotion

bmulholland:
Okay this is good to go now.

rchl:
I don't see any mention of nuxt/emotion. Did you decide against adding it?

bmulholland:
See https://github.com/nuxt-community/sentry-module/discussions/369#discussioncomment-1694390

rchl:
Thanks

PR #390, opened 2022-02-17
fix(deps)!: make @sentry/webpack-plugin a peer dependency

rchl:
BREAKING CHANGE: To use the "publishRelease" option, it's now necessary to manually install the `@sentry/webpack-plugin` package as a dev dependency.

This is in response to Sentry change https://github.com/getsentry/sentry-webpack-plugin/pull/343 which inadvertently made webpack dependencies get installed in production.

### DIFF
--- a/docs/content/en/sentry/options.md
+++ b/docs/content/en/sentry/options.md
@@ -155,6 +155,12 @@ Normally, just setting DSN would be enough.
 
 ### publishRelease
 
+<alert type="info">
+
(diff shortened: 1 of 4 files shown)


rchl:
While Sentry reverted their change, it is still technically more correct to list `@sentry/webpack-plugin` as a peer dependency as it's not required for normal operation of the module and should only really be required on building.

PR #456, opened 2022-11-13
refactor: rewrite with typescript and module-builder

rtibaldo:
Refactored module using nuxt/module-builder (using typescript) and nuxt/kit to begin working towards nuxt 3 compatibility.

Test are semi-broken as module-builder requires nuxt 3 so they now have their own package.json with nuxt 2. 
I don't know if this is the best solution, but I couldn't come up with anything else


rchl:
Haven't looked at the changes yet but are you saying that it should be possible to use module-builder and nuxt/kit and at the same time still have it compatible with Nuxt 2?

rtibaldo:
Yes, we should be able to have compatibility with both nuxt 2 and 3.
The only thing that has a completely different sintax are plugins but since we are using templates for those we can circumvent that. The unocss nuxt module does that and has nuxt 2 and 3 compatibility

src/module.ts
```diff
@@ -0,0 +1,136 @@
...
+export interface ModulePrivateRuntimeConfig extends DeepPartialModuleConfiguration {}
+
+const logger = useLogger('nuxt:sentry')
+
+export default defineNuxtModule<ModuleConfiguration>({
+  meta: {
+    name: '@nuxtjs/sentry',
+    configKey: 'sentry',
+    compatibility: {
+      nuxt: '^2.15.8',
```
rtibaldo, line 23: I didn't know if this module had any nuxt 2 compatibility constraints so I use the last version, we could probably lower that

package.json
```diff
@@ -12,74 +12,68 @@
...
-    "dev:generate": "nuxt generate -c ./test/fixture/default/nuxt.config.js --force-build",
-    "lint": "eslint --ext .vue,.js,.ts lib test types",
+    "lint": "eslint --ext .vue,.js,.ts src test types",
     "lint:fixture": "eslint --ext .vue,.js --no-ignore 'test/fixture/*/.nuxt/sentry.*'",
     "release": "release-it",
-    "test:fixture": "jest --runInBand",
-    "test": "yarn lint && yarn test:fixture && yarn lint:fixture",
+    "test:fixture": "NODE_OPTIONS=--experimental-vm-modules jest --runInBand",
+    "test": "yarn lint && yarn prepack && yarn test:fixture && yarn lint:fixture",
     "coverage": "codecov"
```
rchl, line 42: Running `prepack` depends on `dev:prepare` due to `"extends": "./playground/.nuxt/tsconfig.json"` in `tsconfig.json`.
Running `dev:prepare` depends on `prepack` since it requires module to be built first.
So it's quite a conundrum.

Also, why is `prepack` not called `build`?

src/module.ts
```diff
@@ -0,0 +1,136 @@
...
+      options.publishRelease = defu(options.publishRelease, defaultsPublishRelease)
+    }
+
+    if (serverSentryEnabled(options)) {
+      // @ts-ignore
+      nuxt.hook('render:setupMiddleware', app => app.use(SentryHandlers.requestHandler(options.requestHandlerConfig)))
+      // @ts-ignore
+      nuxt.hook('render:errorMiddleware', app => app.use(SentryHandlers.errorHandler()))
+      // @ts-ignore
+      nuxt.hook('generate:routeFailed', ({ route, errors }) => {
```
rchl, line 82: What's the deal with those hooks in Nuxt 3? Are those flat-out not supported (in which case how would that code work?) or not officially supported but still working? Or does this only work in Nuxt 2?

(Type ignoring is obviously not ideal so I would like to know the reason for that)

build.config.ts
```diff
@@ -0,0 +1,25 @@
...
+    'webpack',
+    '@sentry/tracing',
+    '@sentry/types',
+    '@sentry/browser',
+    '@sentry/node',
+    '@sentry/cli',
+    '@sentry/minimal',
+    'vuex',
+    '@nuxt/types',
+  ],
```
rchl, line 24: Since `unbuild` documentation is lacking, can you educate me on why this is needed?

package.json
rtibaldo: I agree build would have more meaning, I left it as prepack as that was how it was called in the module builder template.
And no dev:prepare doesn't depend on prepack

src/module.ts
rtibaldo: First of all this is just the first step to make it compatible with Nuxt 3 it is not right now. And as for my understanding of the hooks situation I think it is as you said and they only work on Nuxt 2 and Nuxt 3 just ignores them.

The type ignoring is due to using Nuxt 3 types and those hooks not being present for Nuxt 3 🙈

build.config.ts
rtibaldo: So I'll be honest and just say that I'm in the same boat. My understanding is that it should be something akin to the types on a tsconfig and are used during the build process for typechecking but I'm not sure.
I just kept adding untill there were no more errors in the build

src/module.ts
rchl: If this is only the first step then I guess it would make sense to merge it into a `next` branch or something rather than to `main`. Or do you mean to continue working on it in this PR?

package.json
rchl: > And no dev:prepare doesn't depend on prepack

I'm not sure if I was clear. I meant that there is a cyclic dependency between those two. One requires the other to be run first for itself to work. So this kinda impossible situation unless one creates the `./playground/.nuxt` first by some means (by removing dependency on the Sentry module from the playground, for example).

.eslintrc
```diff
@@ -0,0 +1,31 @@
+{
+  "extends": [
+    "@nuxtjs/eslint-config-typescript"
```
rchl, line 3: I actually prefer using `.eslintrc.js` variant since it can be eslint-ed itself.

Also, I have been explicitly avoiding using `@nuxtjs/eslint-config-typescript` since it has some design issues (https://github.com/nuxt/eslint-config/issues/82).

rchl: Actually `@nuxtjs/eslint-config-typescript` was used before too but it was scoped to .ts files to partially address the issues I have with it. If there is no good reason for this change then let's revert to the old approach.

types/sentry.d.ts
```diff
@@ -48,31 +48,35 @@ export interface TracingConfiguration {
...
+  /** See available options at https://github.com/getsentry/sentry-webpack-plugin */
+  publishRelease: boolean | SentryCliPluginOptions
+  runtimeConfigKey: string
+  serverConfig: SentryOptions
+  serverIntegrations: IntegrationsConfiguration
+  sourceMapStyle: WebpackOptions['devtool']
+  requestHandlerConfig: Handlers.RequestHandlerOptions
 }
+
+export interface DeepPartialModuleConfiguration extends DeepPartial<ModuleConfiguration> {}
```
rchl, line 82: Can you explain the use of `DeepPartial` here rather than just `Partial`? I would think that only the root properties should be made optional. Once the root property is provided then the values should be provided per original definition.

Unless there are some specific exception that I'm aware of.

rchl: Well, it looks like it it's OK to do DeepPartial right now, at least.

package.json
rchl: Actually tried it again and `dev:prepare` does seem to complete successfully.
Maybe let's then add a `"prepare": "yarn dev:prepare"` script so that this runs automatically on installing dependencies and doesn't require an extra step.

src/module.ts
rtibaldo: A next branch would be better. My goal for this PR was to prepare the path to add the Nuxt 3 compatibility without breaking the Nuxt 2 compatibility

.eslintrc
rtibaldo: I'll revert that than

types/sentry.d.ts
rtibaldo: My idea was if someone were to use some nested options but not all the object like someone wants to add tracing but lower the sample rate 
`{
...
tracing: { tracesSampleRate: 0.4, },
...
}`. 
They would input something like this so the other values would be undefined or am I missing something?

rchl: Yes, that's fine but I would expect in this case that `TracingOptions` properties are already specified as optional.

I'm thinking that it would not be correct in case we have an option like:

```ts
foo?: FooOptions
```

where `FooOptions` has some required properties like:

```ts
needsthis: string
bar?: string
```

In that case we'd want to preserve original types of `FooOptions`.

That said, I don't think there are currently any options that require that pattern but there seem to be options that need the opposite actuallly (like `publishRelease?: boolean | Partial<SentryCliPluginOptions>`).

rtibaldo: The reason I didn't go that way is that after we merge the user values with the default values we alwasy get the complete object. 

The only value I can see we might want to force the user to provide is the authToken if they want to publish the release, all the other values can be defaulted to the default values

.eslintrc
rtibaldo: I reverted the changes to the eslintrc and fixed the issues that came up after that

package.json
rtibaldo: > let's then add a `"prepare": "yarn dev:prepare"`

Nice idea!

src/core/hooks.ts
```diff
@@ -1,74 +1,86 @@
...
 /**
  * Resolves the options and creates the plugins and the templates at build time.
  *
- * @param      {ThisParameterType<import('@nuxt/types').Module>} moduleContainer
- * @param      {import('../../types/sentry').ResolvedModuleConfiguration} moduleOptions The module options
- * @param      {import('consola').Consola} logger The logger
+ * @param      {Nuxt} nuxt
+ * @param      {ModuleConfiguration} moduleOptions The module options
+ * @param      {Consola} logger The logger
  * @return     {Promise<void>}
```
rchl, line 23: We don't need those JSDoc annotations anymore when specifying types inline. The extra descriptions don't provide any extra value anyway.

rtibaldo: Removed in [`b01023c`](https://github.com/nuxt-community/sentry-module/pull/456/commits/b01023cc5b7ddcd125fe4287e3ac882e86267fc3)

build.config.ts
rtibaldo: I found this message from @harlan-zw on the Nuxt discord as some other users was asking about the externals option

> well when you have it as a dev dependency it means the end user won't install it, so your package wouldn't work
> to try and resolve that, nuxt-module-builder (unbuild) presumes that you want to instead inline the dependency (combining the dependencies code with your module code)
> it's really useful when you are depending on a third party package that you only need some of the code from as it tree-shakes anything not needed

Since the only imports from those packages are types I assume unbild is not inlining anything anyway

rchl: Is it possible to globally disable that instead of having to explicitly exclude? I can't imagine any case when I would want to include dev dependency in built output.

.eslintrc.cjs
```diff
@@ -2,7 +2,10 @@ module.exports = {
   extends: [
     '@nuxtjs/eslint-config',
   ],
-  ignorePatterns: ['dist/'],
+  ignorePatterns: [
+    'dist/',
+    'templates/',
+  ],
```
rchl, line 8: Can you also add this in typescript section so that the style matches the old JS code:

```js
        'comma-dangle': 'off',
        '@typescript-eslint/comma-dangle': ['error', 'always-multiline'],
```

?

daniluk4000:
@rtibaldo I've tried to use come of your changes

"context.$sentry.setUser is not a function"

rchl:
> @rtibaldo I've tried to use come of your changes

On which Nuxt version? Those changes do not make module work in Nuxt 3 (at least not all functionality).

daniluk4000:
> > @rtibaldo I've tried to use come of your changes
> 
> On which Nuxt version? Those changes do not make module work in Nuxt 3 (at least not all functionality).

Nuxt 2 + Nuxt Bridge

.eslintrc.cjs
rtibaldo: Done in [`96c68c0`](https://github.com/nuxt-community/sentry-module/pull/456/commits/96c68c02b608eca450021375408defd7e6a07a00)

rtibaldo:
> Nuxt 2 + Nuxt Bridge

My scope for this PR was to rewrite the module using nuxt module-builder without breaking Nuxt 2 functionality and then in a following PR add Nuxt 3 (and Bridge) support.

As for your issue I tried it locally upgrading the default test app to Bridge and I think the issue is on the client side and not on the server as this was the '_test_' result
image

rtibaldo:
> Test are semi-broken as module-builder requires nuxt 3 so they now have their own package.json with nuxt 2.
> I don't know if this is the best solution, but I couldn't come up with anything else

So I've never worked with a monorepo before (and I could be saying something stupid) but wouldn't using one be the best long term solution to handle testing different versions of Nuxt?

Wdyt @rchl?

rtibaldo:
I dug up what the issue was and found a workaround for the problem I was having with Bridge, maybe it would resolve also yours @daniluk4000.

It seems like there is a problem with the esm version of tslib that gives an error when importing on the client side. 
image

I found this [issue on the Nuxt Bridge repo](https://github.com/nuxt/bridge/issues/25) where @danielroe gave a workaround and explained why there is this error.

TL DR: Add this to your nuxt config
```
  alias: {
    tslib: 'tslib/tslib.es6.js'
  },
```

PS. This does not mean the module is Bridge compatible, there are things that I'm pretty sure don't work, like Nitro errors being traced

daniluk4000:
@rtibaldo looks interesting, I'll try that tomorrow, thanks

rchl:
> > Test are semi-broken as module-builder requires nuxt 3 so they now have their own package.json with nuxt 2.
> > I don't know if this is the best solution, but I couldn't come up with anything else
> 
> So I've never worked with a monorepo before (and I could be saying something stupid) but wouldn't using one be the best long term solution to handle testing different versions of Nuxt?

I don't necessarily see a need for monorepo just yet.
I think the main issue with the test setup right now is that the nuxt2 directory extends its tsconfig.json (https://github.com/rtibaldo/sentry-module/blob/daa948ca698a567309e7727d48f76d3e24a616ba/test/tsconfig.json#L2-L2) with the one generated by Nuxt 3 playground and that is not gonna work. It should just have its own, independent config.

BlakeB415:
Nuxt v3 is able to run in Cloudflare Workers. Will this be compatible in that context?

rchl:
I think we're gonna handle upgrade to v7 SDK first (https://github.com/nuxt-community/sentry-module/pull/461), release as new major version and then pick this up. This will also likely have to be released as yet another major version.

manniL:
@rchl as #461 is through, this PR needs adaption I suppose?

rchl:
I did rebase on `main` and made some small changes relative to this PR. Those are in 7fb1c5ea8c3bc6c1cc0631a3afc7b5a6878b1a96

Haven't updated this one because it would be impossible to see changes that I made. We can possibly create new PR from scratch.

That said, I still have doubts about bridge because:
a) Nuxt Bridge is not released in stable version and requires nuxt-edge which makes it pretty much useless for production services
b) I'm not sure if it's possible to replicate this stuff in Nitro: https://github.com/nuxt-community/sentry-module/blob/7fb1c5ea8c3bc6c1cc0631a3afc7b5a6878b1a96/src/module.ts#L110-L126 and there is lack of documentation or help on it

I suppose it would be easy to create a version that would only work in webpack context but Nitro ("server middleware) is a mystery.

rchl:
Merged changes here after all. Merge commit should be fairly readable.

rchl:
I'm not sure why generated file in `/dist/module.mjs` contains absolute local paths. This will obviously break hard when released.

@rtibaldo do you plan to pick this up again at some point?

rtibaldo:
I will try to take a look at it this weekend.

Since I'm here I'll try to give my two cents about your doubts.
I agree with you that bridge not being stable is a problem and that's why I think bridge support should be a consequence and not a goal. Since it shares a lot of features with Nuxt 3, bridge compatibility should almost come by itself once Nuxt 3 compatibility is achieved.

About your Nitro question I don't have much to say other than imo there should be a different package for a Nitro plugin that this module would inject in the config.

manniL:
> I will try to take a look at it this weekend.

Thanks!

> About your Nitro question I don't have much to say other than imo there should be a different package for a Nitro plugin that this module would inject in the config.

Well, ideally this module could offer e.g. a Nitro middleware that you could plug in, either manually or via config. Maybe even how [`nuxt-auth` from sidebase](https://github.com/sidebase/nuxt-auth) is doing. I'm happy to collab on it after the rewrite has been released.

package.json
```diff
@@ -12,56 +12,50 @@
...
+  },
+  "main": "./dist/module.cjs",
+  "types": "./dist/types.d.ts",
+  "files": [
+    "dist",
+    "types"
+  ],
   "scripts": {
+    "prepare": "yarn dev:prepare",
+    "prepack": "nuxt-module-build",
```
rchl, line 33: The `prepare` script causes problems when publishing as it runs after `prepack` but before the artifact is created so it reverts to stubbed build for development.

rchl:
BTW. I've released current state of this code as `@nuxtjs/sentry@beta`.

I seem to be getting:

> Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.

when trying it quickly in a Nuxt 2 project but haven't had time to look closer into it.

rtibaldo:
I should have fixed the es module error and some other things that broke with the refactor

src/core/hooks.ts
```diff
@@ -1,85 +1,78 @@
...
+import type { ModuleConfiguration, SentryHandlerProxy } from '../../types'
+import { clientSentryEnabled, serverSentryEnabled, envToBool, canInitialize } from './utils'
+import { resolveRelease, ResolvedClientOptions, resolveClientOptions, ResolvedServerOptions, resolveServerOptions } from './options'
+
+const RESOLVED_RELEASE_FILENAME = 'sentry.release.config.mjs'
+
+export async function buildHook (nuxt: Nuxt, moduleOptions: ModuleConfiguration, logger: Consola): Promise<void> {
   const release = await resolveRelease(moduleOptions)
 
+  const templateDir = fileURLToPath(new URL('./templates', import.meta.url))
```
rchl, line 21: When `dist` is mocked then this doesn't work because `templates` directory is one directory up from where `hooks.ts` is in. But when `dist` is built then this code is merged into `module.mjs` so then it works.

rchl: move files from `core` to the root.

rchl:
Situation with Nuxt testing is rather messy.

Currently module uses `@nuxtjs/module-test-utils` which is not compatible with Nuxt 3.
There is `@nuxt/test-utils v0.x` which is only compatible with Nuxt 2.
There is also `@nuxt/test-utils v3.x` that is in theory compatible with both Nuxt 2 and Nuxt 3 (not 100% sure about Nuxt 2) when using `jest` runner (rather than `vitest`) but it [doesn't really work with jest](https://github.com/nuxt/nuxt/issues/18436).

package.json
lihbr: Just sharing some (late) context, the Nuxt 3 module template uses the `prepack` script hook to be sure the module gets build before `npm publish` (since a build step is now _mandatory_ to build modules when it wasn't with Nuxt 2 module flavor)

lihbr:
Hey there, I had a look at the project here and attempted to find a way to get the test setup to work in a smooth manner.

_(tl;dr;)_ So far, unfortunately, I haven't found any compelling way to test a Nuxt module against both Nuxt 2 & Nuxt 3.

Nonetheless, I figured out it might be interesting to share some findings/potential workarounds given the context here.

## `@nuxt/test-utils@3` status

`@nuxt/test-utils@3` is meant to support Nuxt 3 and Nuxt Bridge, ideally, we'd like also for it to support Nuxt 2. It does to some extent, but roadblocks are quickly encountered as mentioned by @rchl.

The journey for `@nuxt/test-utils@3` to also support Nuxt 2 is not that straightforward unfortunately. I noted some of the issues I was facing in this document if you want to have a look at it (temporary): https://lihbr.notion.site/lihbr/nuxt-test-utils-exploration-1d6345640285449897cff2176f7bf53a

## Potential workarounds

### 1. `nuxt2-test` and `test` folders

This one could be quite low-effort to get up and running with. Similarly to `nuxt2-playground`, we could have a `nuxt2-test` with its own `package.json` and have tests written in it with `@nuxtjs/module-test-utils` (they way tests are written today in `test`)

Nuxt 3 tests would be available under `test`, and would be written using `@nuxt/test-utils@3`. Those would be about translating the existing tests written with `@nuxtjs/module-test-utils` to `@nuxt/test-utils@3` syntax, but this should be pretty straightforward.

### 2. Strengthen unit test strategy and (optionally if not doing n°1) drop Nuxt 2 e2e tests

We could make sure the module behaves as expected depending on its context by mocking its input and spying on its output.

That's something I did with the Prismic module here: https://github.com/nuxt-modules/prismic/blob/v3/test/module-options.test.ts#L10-L35 Nuxt doesn't provide yet those APIs so I wrote them as test utils there. I also shared an RFC about providing unit-testing utils there: https://github.com/nuxt/nuxt/discussions/18399

### 3. Make a new major only targeting Nuxt 3

In all fairness, I'm not sure how realistic that is for this module as it looks like you're still iterating regularly on its Nuxt 2 version and v7 is pretty recent, but let's mention it anyway!

To share my own experience, I did so for the Prismic module as it sounded just simpler to me this way. As a result of this, I've been juggling with maintaining two versions for some time now, and this has been pretty smooth so far (in my case the Nuxt 2 version of it can be considered in "maintenance" mode, I'm not looking to add new features to it anymore)

---

I'm conscious none of the workarounds above are really exciting, but that's what I figured out for now 🤷‍♀️
In any case, I hope this helps, at least a bit ☺️

rchl:
Appreciate your help and looking into it.

From my experimentation before I've kinda concluded that there is too much issues and it's not really feasible to test both. Duplicating tests would probably make it work but it's missing the point which for me is having to only maintain one version. Same for having a separate major version for Nuxt 3.

What is the most disappointing for me is that the whole Bridge and supposed interoperability between Nuxt 2 and Nuxt 3 is not here at all and probably never will be and I would feel some consolation if Nuxt authors would admit that instead of pretending that it's a thing. Sorry for this little rant from me but I'm a bit frustrated. :)

We could remove all traces of Nuxt 3 from this PR and just make this about using `@nuxt/module-builder` and rewriting in typescript which has a value in itself but it won't really get us closer to supporting bridge or Nuxt 3.

src/hooks.ts
```diff
@@ -1,85 +1,78 @@
...
-    moduleContainer.options.build.plugins = moduleContainer.options.build.plugins || []
-    moduleContainer.options.build.plugins.push(
-      new webpack.DefinePlugin({
-        __SENTRY_DEBUG__: 'false',
-      }),
-    )
+    addWebpackPlugin(new webpack.DefinePlugin({
+      __SENTRY_DEBUG__: 'false',
+    }))
+    // TODO: Handle Vite
```
rchl, line 54: The `addWebpackPlugin` helper [uses `webpack:config` hook](https://github.com/nuxt/nuxt/blob/64b835c70701c8f27debe8a534ec99034bda812b/packages/kit/src/build.ts#L43-L56) running it here might be too late.

lihbr:
You're welcome, and no worries, I understand the frustration~

I'll let you know if I find anything that could make it work in an integrated way.

daniluk4000:
@rchl sorry to ask, but when will moment come for you to support Nuxt 3 only? What's the reason to drop newer version because you can't support two versions at once? 

P.S. Nuxt team will probably abandon Bridge and integrate some functions in Nuxt 2 itself, as one of devs told me. I totally understand frustration as well because I didn't succeed with setting up Bridge as well, and I had to make a fork & PR (https://github.com/nuxt-community/composition-api/pull/742) to Nuxt CAPI in order to even use it. 

It seems for me that it's better to drop Nuxt 2 & Bridge and migrate directly to Nuxt 3 as it seems pretty much stable now and becomes more stable day by day. Nuxt team will not change API anymore until Nuxt 4 and will follow semver.

rchl:
I suppose we'll adjust this PR to only support and test Nuxt 2 and then it might be necessary to create a Nuxt 3 specific version.

I don't personally use Nuxt 3 so I'm not really that motivated to work on it so hopefully someone can help with that.

daniluk4000:
@rchl offtop - when do you plan to migrate (if you have plans at all)? :D

Speaking of our topic, we plan to migrate in next few month or so (doing preparations now). I guess I'll be pretty motivated to create Nuxt 3 version (or at least help with that), but I believe that this PR must be merged before that.

rchl:
> @rchl offtop - when do you plan to migrate (if you have plans at all)? :D

nowhere on the horizon yet. There are still many modules (including less popular) that don't support Nuxt 3.

rchl:
I've released the current state as `@nuxtjs/sentry@beta` (`@nuxtjs/sentry@8.0.0-beta.2`).

Testing likes to crash for some reason on CI. Works locally (usually).

src/hooks.ts
rchl: This works fine because it's done from `before:build` hook.

rchl:
A bit later but just noticed another deal breaker. `@nuxt/kit` depends on `@nuxt/schema` which [depends on vue 3](https://github.com/nuxt/nuxt/blob/71225e50c5bc45c3feb7b189997659d880fe03f3/packages/schema/src/types/config.ts#L1-L1) which completely breaks types when using beta release of this module in a Nuxt 2 project...

rchl:
> the issue I think you might be encountering is this one: [nuxt/nuxt#19526](https://github.com/nuxt/nuxt/pull/19526).

That's one of the issues that we've discussed in private.

The other issue has to do with using `@nuxt/module-builder` which automatically generates module schema for the module. That results in `@nuxt/schema` being referenced in the resulting type definitions (`module.d.ts` in our case).

Specifically this line:

```
import * as _nuxt_schema from '@nuxt/schema';
```

You've provided a possible workaround by casting module export to `any` like:

```ts
export default defineNuxtModule(...) as any
```

and it does in fact get rid of `@nuxt/schema` references but I'm not gonna believe that this is the official way to do it. It kinda defeats the purpose of using `@nuxt/module-builder` in the first place. I could instead use some other typescript compiler unrelated to Nuxt.

(The `@nuxt/module-builder` is advertised as "Compatible with Nuxt 3 and Nuxt Kit".)

> Just checking, @nuxt/kit and @nuxt/schema do not depend on Vue 3 at all. 

Maybe our understanding of "depends on" differs because those `vue` types are definitively not in Vue 2 (not even in 2.7):

https://github.com/nuxt/nuxt/blob/71225e50c5bc45c3feb7b189997659d880fe03f3/packages/schema/src/types/config.ts#L1-L1

rchl:
I've missed before that https://github.com/nuxt/module-builder/blob/main/README.md says "Nuxt 3 or Nuxt Bridge. Nuxt 2 is functional but not advised". So I guess it's not lying. The module will work but the types will be broken in Nuxt 2. This could perhaps be more clear.

rchl:
And even if I can workaround it for the published package, the types (like extends of `NuxtConfig`) will still be broken when locally developing this project because I can't avoid referencing `@nuxt/schema` (through `@nuxt/kit`) there.

rchl:
I think the next step is to replace `nuxt-module-builder with `unbuild` or plain rollup since I don't want to loose types while developing the module (even if it only affects test fixtures).

rchl:
I don't have any outstanding issues with those changes so released as `@nuxtjs/sentry@8.0.0-beta.8` (`beta` tag) and will probably merge this soon once I test it a little.

rchl:
I don't consider this having any breaking changes (maybe only some minor in types but shouldn't affect normal usage) so I will aim to release it as a minor or patch bump.

Issue #433, opened 2022-07-12
Upgrade to Sentry packages v7


bmulholland:
https://github.com/getsentry/sentry-javascript/releases/tag/7.0.0

alexcroox:
Sentry blogged about the size reduction in v7, it would be great to have this updated so we can improve the perf of our Nuxt apps.

https://blog.sentry.io/2022/07/19/javascript-sdk-package-reduced/

@rchl is there a possibility of having Sentry as a peer dependancy so this module doesn't need to be re-published every time they make a release?

rchl:
> @rchl is there a possibility of having Sentry as a peer dependancy so this module doesn't need to be re-published every time they make a release?

`npm` (and `yarn` too, pretty sure) will update the `@sentry/*` depenencies within specified range when running `npm update @nuxtjs/sentry` so there is not really a need to make those into peer dependencies.

bmulholland:
The range specifies 6, so upgrading to 7 is not possible right now.

rchl:
My answer was not in relation to updating to version 7 but only about upgrading to new patch/minor releases within the v6 range.

As far as upgrading to v7, this is a task on its own as it needs changes in the module and using peer dependencies wouldn't change anything in that regard.

alexcroox:
good to know about running npm update for minor bumps, thanks

iTechSebastianoLaini:
any ETA on this? @sentry/tracing it's blocked to V6 because @nuxtjs/sentry still used V6 and not 7.

glebv:
Hi, our company is going to use sentry-module in production but it doesn't support v7 libs. sentry.io ui warns that our code uses old sdk. Could you share some estimates or insights about the migration? Should we expect on the soon migration? 
Thanks a lot for your work!

rchl:
Not making any promises due to short on time.

Technically there should be no issues with using v6. We are doing that at my company and as long as it works fine it's not a priority to update.

glebv:
@rchl thanks for your answer. Got it.

vedmant:
I'm currently optimizing my nuxt app for google page speed tests and sentry code is quite a large piece of code. Upgrading to version 7 would be really helpful. Otherwise looks like I need to copy paste plugins and use sentry package directly.

bmulholland:
@rchl Do you have time to post a short description of what would be needed to make the upgrade happen? That way, maybe someone for whom it's a higher priority could make an attempt.

[...]

Issue #417, opened 2022-05-07
Use @sentry/vue instead of browser with vue integration


pimlie:
### Is your feature request related to a problem? Please describe.

No problem, just small optimization

### Describe the solution you'd like

Use `@sentry/vue` instead of `@sentry/browser`

### Describe alternatives you've considered

na

### Additional context

Seems a bit simpler solution as we wouldnt have to split browser & vue options anymore

```js
// @see https://docs.sentry.io/platforms/javascript/guides/vue/configuration/integrations/vue-router/
Sentry.init({
  Vue,
  trackComponents: true,
});
```

rchl:
Hmm, I've read that documentation and I'm confused as it doesn't seem to be saying anything to the effect of the this issue.

I do agree that we should use `@sentry/vue` for better VueRouter integration for tracing but that doesn't seem to be in any way connected to dropping the Vue integration.

- That documentation page seems to be talking only about the tracing part and never mentions the Vue integration
- The `Vue` in the configuration snippet that you've pasted refers to the `Vue` itself and not the `Vue` Sentry integration
- That snippets also seems to be missing `integrations: [new BrowserTracing()],` line (which to me looks like an omission on their part

pimlie:
See the package readme for more info: https://github.com/getsentry/sentry-javascript/tree/master/packages/vue

Its just a wrapper around @sentry/browser specifically for Vue.

rchl:
Ah, so it provides own `Sentry.init` that initializes the Vue integration automatically. That wasn't clear from that documentation page.

The Vue integration options also become part of the `init` options then.

I don't mind changing to it. More future proof, I guess.

pimlie:
I can share my updated lazy template if you wish? Im not planning to update the options of the module thought as that would probably be a breaking change

rchl:
I'm not sure what I would do with it if we are also not updating the settings at the same time :)

The current master version already has breaking changes that weren't released yet. For example removal of deprecated `attachCommits`, `repo` and `webpackConfig` options. Also refactorings that I wasn't entirely sure about so didn't commit to a release yet...

So it would be a fairly good moment to do breaking changes now and release new major version.
The only thing is that Sentry seems to be planning a major release soon also so maybe we should wait for that.

rchl:
Sentry 7 actually removes the `Vue` integration from `@sentry/integrations` so doing this makes even more sense. But it would make sense to do that as part of upgrading to Sentry 7 when final version is released.

[...]

PR #461, opened 2022-12-01
feat!: update Sentry SDK from v6 to v7

LukasHirt:
## Summary
- update all sentry packages
- drop sentry/browser
- add sentry/vue
- drop Vue from integrations
- get types from sentry/core instead of deprecated sentry/minimal

## Related issues
- closes https://github.com/nuxt-community/sentry-module/issues/417
- closes https://github.com/nuxt-community/sentry-module/issues/433


LukasHirt:
@rchl could you pls look into reviewing this PR? It's quite important for our project to get sentry to v7. I'm happy to take care of any changes necessary...

I run the tests locally and they passed and also tried to link it to our project and reported mock errors from there. I went through the docs and tried to spot any reference that needs to be updated.

LukasHirt:
Hmm, I missed some parts in the options.js file still using sentry/browser - will fix that

LukasHirt:
@rchl ok, now it should be ready for review

rchl:
I suppose this requires new major version of the module and adding migration guide (like https://i18n.nuxtjs.org/migrating).

Also there is a question of whether this should have its own major version and then https://github.com/nuxt-community/sentry-module/pull/456 should trigger yet another major version or whether those two changes should be sharing same, new major version (thus requiring nuxt/kit and potentially causing extra issues for Nuxt 2 users...).

lib/core/options.js
```diff
@@ -96,16 +96,10 @@ function resolveTracingOptions (tracing, config) {
...
-          hooks: ['mount', 'update'],
-          timeout: 2000,
-          trackComponents: true,
-        },
-      },
-      browserOptions: {},
+      hooks: ['mount', 'update'],
+      timeout: 2000,
+      trackComponents: true,
+      tracePropagationTargets: ['localhost', /^\//],
```
rchl, line 102: This seems to be the default value. Is it necessary to define it?

And related to that, why are we only passing `tracePropagationTargets` to `new BrowserTracing`? Shouldn't we allow the user to override all the options that it supports?

darthf1:
> Also there is a question of whether this should have its own major version and then #456 should trigger yet another major version or whether those two changes should be sharing same, new major version (thus requiring nuxt/kit and potentially causing extra issues for Nuxt 2 users...).

Does #456 for Nuxt 2 require a dependency on `nuxt/bridge`? If so:

If possible, and if they can be done seperately, I would prefer two seperate majors with the upgrade to Sentry 7 before the #456. This means existing users can upgrade, without having to move to either `nuxt/bridge` or `nuxt 3`.

`nuxt/bridge` still requires a dependency on `nuxt-edge`, and migration to Nuxt 3 is proving to be quite painful for me. So it would be great if the upgrade to `Sentry 7` would be possible without having to touch anything else.

rchl:
Yes, i agree with that given that migration to bridge can prove problematic

lib/plugin.client.js
```diff
@@ -54,7 +48,10 @@ export default async function (ctx, inject) {
     }).join(',\n    ')%>,
   ]
   <% if (options.tracing) { %>
-    config.integrations.push(<%= `new TracingIntegrations.BrowserTracing(${serialize(options.tracing.browserOptions)})` %>)
+    const { tracePropagationTargets, ...tracingOptions } = options.tracing
+
+    config.integrations.push(new BrowserTracing({ tracePropagationTargets }))
+    merge(config, tracingOptions)
```
rchl, line 54: `options` is an undefined variable. Reproduces with tracing enabled.

lib/core/options.js
```diff
@@ -96,16 +96,10 @@ function resolveTracingOptions (tracing, config) {
   /** @type {NonNullable<import('../../types/sentry').TracingConfiguration>} */
   const tracingOptions = merge(
     {
-      tracesSampleRate: 1.0,
```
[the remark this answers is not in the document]
rchl: I've re-added it. Let me know if you disagree with that.

That said, I'm also wondering whether we have to include the defaults at all. Maybe we should just not provide a configuration at all (when using `tracing: true`) and let Sentry use its defaults.

LukasHirt:
@rchl thanks for your review and your commits! I'll try to get this week the time to look into this again.

rchl:
I've picked up some work on this module and created various commits on master that can conflict with this task. I can help with merging and resolving conflicts.

lib/core/options.js
LukasHirt: You're right. It's now possible by passing the options into `browserTracing` key in the tracing options.

lib/core/options.js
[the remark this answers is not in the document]
LukasHirt: > I've re-added it. Let me know if you disagree with that.

I originally moved it out of the tracing object as I confused myself with the structure and then forgot to bring it back 🙈 THX for handling that.

> That said, I'm also wondering whether we have to include the defaults at all. Maybe we should just not provide a configuration at all (when using tracing: true) and let Sentry use its defaults.

I think that won't work for 100%. It looks like Sentry is not using defaults for all the values. Specifically the `tracesSampleRate` seems to be missing in defaults.

types/sentry.d.ts
```diff
@@ -44,19 +39,11 @@ interface TracingOptions {
...
-}
-
-export interface TracingConfiguration {
-    tracesSampleRate?: number
-    vueOptions?: {
-        tracing?: boolean
-        tracingOptions?: Partial<TracingOptions>
-    }
-    browserOptions?: Partial<BrowserTracing['options']>
+    tracePropagationTargets: (string|RegExp)[]
```
rchl, line 42: Various issues with the `TracingConfiguration` interface.
- properties should be optional
- `browserTracing` is missing
- `tracePropagationTargets` belongs in `browserTracing`

types/sentry.d.ts
```diff
@@ -1,9 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { IncomingMessage, ServerResponse } from 'http'
 import { Options as WebpackOptions } from 'webpack'
-import { BrowserTracing } from '@sentry/tracing'
 import { Options as SentryOptions } from '@sentry/types'
-import { BrowserOptions } from '@sentry/browser'
+import type { Options as SentryVueOptions } from '@sentry/vue/types/types'
```
rchl, line 5: No need to use `import type` inside a type definition file really.

rchl: addressed in f6e67a8f39107583ea52da6e90dfd17158789e81

types/sentry.d.ts
rchl: addressed in f6e67a8f39107583ea52da6e90dfd17158789e81

lib/core/options.js
```diff
@@ -97,16 +97,9 @@ function resolveTracingOptions (options, config) {
...
-        tracingOptions: {
-          hooks: ['mount', 'update'],
-          timeout: 2000,
-          trackComponents: true,
-        },
-      },
-      browserOptions: {},
+      tracesSampleRate: 1,
+      trackComponents: true,
+      browserTracing: {},
```
rchl, line 102: I'm thinking now that the `vueOptions` should still be separated since the tracing options are also used on the server-side while options like `trackComponents` and `hooks` are Vue-only.

rchl: Addressed in 19357615abab42136c488e459bb4ba62eded8217

rchl:
Apart from tweaking the docs, I'm pretty happy with the changes and ready to merge when you are.

I will still likely do some follow up changes to enable vue integration and support tracing in `lazy` mode but that's for another PR.

rchl:
Thanks!

PR #514, opened 2023-02-28
fix(tracing): automatically instrument server-side requests

### LINKED ISSUE #500: Is perfomence monitoring really enabled on server?
[Documentation ](https://sentry.nuxtjs.org/sentry/options/#tracing) says that `tracing` option
> Enables Sentry Performance Monitoring on the [server](https://docs.sentry.io/platforms/node/performance/) and [browser](https://docs.sentry.io/platforms/javascript/guides/vue/performance/) side.

but I don't see any transaction from server side in sentry...

In sources I found [code](https://github.com/nuxt-community/sentry-module/blob/9614795af3a5aa9730a2fbf4346f1eb0f7010059/lib/core/hooks.js#L169-L170) which adds errorHandler and requestHandler, but [sentry documentation](https://docs.sentry.io/platforms/node/guides/express/performance/) for nodejs perfomence says we should use tracingHandler after requestHandler

Also [there is ability](https://docs.sentry.io/platforms/javascript/performance/connect-services/#pageload) to connect ssr with browser transaction throgh meta tags

rchl:
Fixes #500

### DIFF
--- a/lib/core/hooks.js
+++ b/lib/core/hooks.js
@@ -168,6 +168,9 @@ export async function initializeServerSentry (moduleContainer, moduleOptions, se
     Sentry.init(config)
     sentryHandlerProxy.errorHandler = Sentry.Handlers.errorHandler()
     sentryHandlerProxy.requestHandler = Sentry.Handlers.requestHandler(moduleOptions.requestHandlerConfig)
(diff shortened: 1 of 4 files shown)


arkhamvm:
Great! When it will be available as new release?

PR #520, opened 2023-03-17
Removed misleading config child

HapLifeMan:
When reading the doc, it was unclear if it was necessary to give options into a `config` object.
After testing, the object is not necessary to be able to have it working properly.
All options should be at the root of the module configuration object.

### DIFF
--- a/docs/content/en/guide/setup.md
+++ b/docs/content/en/guide/setup.md
@@ -40,12 +40,10 @@ Then, add `@nuxtjs/sentry` to the `modules` section of `nuxt.config.js` and set 
   sentry: {
     dsn: '', // Enter your project's DSN.
     // Additional Module Options.
-    config: {
-      // Optional Sentry SDK configuration.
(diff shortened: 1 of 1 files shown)


rchl:
Module options should go into the root `sentry` object.

Sentry SDK-specific options should go into the `config` (`clientConfig`/`serverConfig`) objects.

I guess it's easy to get confused about Sentry SDK vs. Sentry Module options but this change doesn't make it clearer.

rchl:
Currently the distinction between between the module options and the SDK options is expressed by stating either "module" or "SDK". SDK is what the module wraps to provide the Sentry functionality in Nuxt.

Maybe renaming option keys would make things clearer (for example `config` => `sentrySdkConfig`)? Feel free to suggest improvements that would make it less confusing.

rchl:
I've updated documentation at ca22ddada5d3f6d1abd55cd4ccaca2a7018dd1b2. Hopefully that's clearer.

PR #550, opened 2023-04-05
chore(ci): add bundle size checks as PR comments

rchl:
Adding CI workflow that will analyze client bundle size and post results in the PR.


rchl:
Will fail because `main` doesn't contain this code yet.

PR #561, opened 2023-04-19
chore(docs): update typos in Session Replay page

ryan-cahill:
Updated documentation to use `Replay` as the client integration key, not `Reply`


rchl:
Thanks!

PR #614, opened 2023-08-27
feat(tracing): allow configuring vueRouterInstrumentation options

### LINKED ISSUE #613: Make transaction name configurable for performance monitoring.
### Is your feature request related to a problem? Please describe.

In Sentry's performance monitoring, I'd like to get the path instead of the route name in Sentry.

### Describe the solution you'd like

`Sentry.vueRouterInstrumentation` actually [supports this](https://docs.sentry.io/platforms/javascript/guides/vue/configuration/integrations/vue-router/#configuration), by passing a configuration object with `routeLabel` to it.

### Describe alternatives you've considered

I've tried using a script through `clientConfig` that is executed on runtime, but I can't figure out how to access the Vue integration through `Context.$sentry` so I could patch it.

### Additional context

Maybe another `vueRouterOptions` parameter would do it? Would then be applied as second parameter in `vueRouterInstrumentation()` [here](https://github.com/nuxt-community/sentry-module/blob/main/src/templates/client.shared.js#L61).

rchl:
Allows setting `tracing.vueRouterInstrumentationOptions` to specify options passed to `vueRouterInstrumentation` integration.

Resolves #613

### DIFF
--- a/docs/content/en/configuration/options.md
+++ b/docs/content/en/configuration/options.md
@@ -317,9 +317,13 @@ export default function () {
     vueOptions: {
       trackComponents: true,
     },
+    vueRouterInstrumentationOptions: {
+      routeLabel: 'name',
+    },
   }
   ```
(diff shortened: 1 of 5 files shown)


freezy:
Awesome!

PR #648, opened 2023-11-18
fix: don't enable RewriteFrames integration by default

rchl:
After inspecting some crashes manually I've realized that the `RewriteFrames` integration that the module enables by default does more harm than good.

In case there are source maps it doesn't really make any difference since stack frame paths are resolved from source maps then.

In case there are no source maps (for example when crash is within the server code and not inside the webpack context) then what `RewriteFrames` does is it [strips the whole path leaving just the file name](https://github.com/getsentry/sentry-javascript/blob/45692f4e80c49bf8aca77695e12313ee24ca8bae/packages/integrations/src/rewriteframes.ts#L83-L83). That can result in just `index.js` being shown, for example, which makes it unclear which file is being referred to.

Decided to not enable this integration by default as a fix rather than a breaking change.

It can easily be re-added manually and ideally with the `root` option specified to just strip the app's root path. And maybe shouldn't be added on the client side since on the client side the source maps should make it redundant.

### DIFF
--- a/docs/content/en/configuration/options.md
+++ b/docs/content/en/configuration/options.md
@@ -220,7 +220,6 @@ sentry: {
   {
     ExtraErrorData: {},
     ReportingObserver: { types: ['crash'] },
-    RewriteFrames: {},
   }
   ```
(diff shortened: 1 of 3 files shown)


rchl:
Well, since the philosophy of Nuxt is to provide good defaults, I suppose I can set the `root` option to a good default value on the server instead...

